PR #124, opened 2025-02-11
versioning upper bounds

Nathanjp91:
Currently if you try to install sdk via poetry after createing a default poetry project, it will error with a maximum version <4.0 on python constraint. This is beacuse of the carat operator implicitly setting an upper bound of next major version. Poetry however creates python projects with >= operator, which can extend beyond major version updates. Minor solution to set a manual upper bound that we'll have to update periodically as python version advances. 

Also tried setting the python version of the project to just be >=3.10, however this issue is present in one of our sub-dependencies as well and we get the same error on installing their package. For now just set upper bound on package to make installing smoother.


robbie-portia (approved):
Good find! Now we just need to write the agent that can auto-increment this for us when new python versions come out...

PR #126, opened 2025-02-12
Add linting to contributing docs

robbie-portia:
* Added a line about linting to the contributing docs
* Ran the linter over all files in the repo which gave a couple of changes

I tried setting up pyright through pre-commit too (using https://github.com/RobertCraigie/pyright-python) but unfortunately it doesn't really work because it can't (easily) pick up the python virtual environment set up by poetry.


TomSPortia (approved):
Tidy 🧹

PR #130, opened 2025-02-14
Fix parameter

TomSPortia:


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "portia-sdk-python"
-version = "0.0.1-alpha.52"
+version = "0.0.1-alpha.53"
```
robbie-portia, line 3: I committed something similar this morning but didn't increment the version. Was this something I should have done?

TomSPortia: Incrementing the version triggers the release to PyPI. We don't have any rules as to what constitutes a release so its not an issue you didn't but if you want people to get the new version you'll need to 😄

PR #131, opened 2025-02-14
Replace instructor with Langchain

TomSPortia:


TomSPortia:
This is the sort of PR it would be amazing to be able to eval before merging

emmaportia:
> This is the sort of PR it would be amazing to be able to eval before merging

You could eval it before merging though?

TomSPortia:
> > This is the sort of PR it would be amazing to be able to eval before merging
> 
> You could eval it before merging though?

Was thinking in an automated manner, but will run it manually!

PR #132, opened 2025-02-14
[POR-713] Initialise llmtools in the plans to ensure it's wrapped wit…

OmarEl-Mohandes:
### Issues

- Steps that are executed by `LLMTool` doesn't get stored in the cloud/dashboard. This is because we don't wrap the call with `ToolWrapper`
- Planner sometimes producing plans with `LLMTool` and sometimes not which is affecting Evals.
- I faced an error before where planner is hallucinating tool_ids,  because we don't check that tools produced by the planner are actually contained in the `tool_list` that was given.

### Fix
- Verify the tool_ids are existent in the tool list, in the planner stage.
- Refill the steps that doesn't contain tool ids with  `LLMTool`.
- Fallback to the LLMTool (just in case) in the runner instead of delegating this decision to verifier agent. 
   - We don't expect this to happen, but just in case to be safe.
   - Raise exception in the verifier agent if no tool found on `execute_sync`

### DIFF
--- a/portia/planners/one_shot_planner.py
+++ b/portia/planners/one_shot_planner.py
@@ -7,7 +7,7 @@ from typing import TYPE_CHECKING
 
 from portia.execution_context import ExecutionContext, get_execution_context
 from portia.llm_wrapper import LLMWrapper
-from portia.open_source_tools.llm_tool import LLM_TOOL_ID
+from portia.open_source_tools.llm_tool import LLMTool
 from portia.planners.context import render_prompt_insert_defaults
 from portia.planners.planner import Planner, StepsOrError
 
@@ -67,7 +67,7 @@ class OneShotPlanner(Planner):
         # Add LLMTool to the steps that don't have a tool_id.
         for step in response.steps:
             if step.tool_id is None:
-                step.tool_id = LLM_TOOL_ID
+                step.tool_id = LLMTool.LLM_TOOL_ID
 
         return StepsOrError(
             steps=response.steps,
--- a/portia/planners/one_shot_planner.py
+++ b/portia/planners/one_shot_planner.py
@@ -7,12 +7,13 @@ from typing import TYPE_CHECKING
 
 from portia.execution_context import ExecutionContext, get_execution_context
 from portia.llm_wrapper import LLMWrapper
+from portia.open_source_tools.llm_tool import LLM_TOOL_ID
 from portia.planners.context import render_prompt_insert_defaults
 from portia.planners.planner import Planner, StepsOrError
 
 if TYPE_CHECKING:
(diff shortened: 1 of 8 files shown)


portia/planners/one_shot_planner.py
```diff
@@ -59,7 +60,40 @@ def generate_steps_or_error(
...
+        for step in response.steps:
+            if step.tool_id is None:
+                step.tool_id = LLMTool.LLM_TOOL_ID
+
         return StepsOrError(
             steps=response.steps,
             error=response.error,
         )
+
+    def _validate_tools_in_response(self, steps: list[Step], tool_list: list[Tool]) -> str | None:
```
TomSPortia, line 77: Is this to stop tool name hallucination? I'm not convinced we should add a point case for this given I don't see it happening regularly? If we want to address that I think we should look at developing a more complicated planner akin to the verifier agent.

portia/runner.py
```diff
@@ -535,9 +536,9 @@ def _handle_clarifications(self, workflow: Workflow, step_output: Output, plan:
         return False
 
     def _get_tool_for_step(self, step: Step, workflow: Workflow) -> Tool | None:
-        if not step.tool_id:
-            return None
-        child_tool = self.tool_registry.get_tool(step.tool_id)
+        # We don't expect to have a none tool_id for a step,
+        # but if we do, we use the LLM tool
+        child_tool = self.tool_registry.get_tool(step.tool_id) if step.tool_id else LLMTool()
```
TomSPortia, line 541: We're defaulting to the LLMTool() in two places right? Here and in the planner. I think we should remove one of the cases. Either the planner should always return a tool or the runner should default if its none but having the behaviour in two places doesn't seem optimal to me.

portia/planners/one_shot_planner.py
OmarEl-Mohandes: I definitely saw it happens more than once with me. Can't reproduce it that much tbh. I agree with your sentiment, but I'd think we should be more on defensive side, rather than trusting llms. It's almost the same case as if we validate the pedantic model from the llm response.

TomSPortia: IMO its the same distinction we have in the agent. The one shot agent also doesn't catch hallucinations, the verifier agent does. I just think you can add many sorts of bandaid fixes to this but we're better off fixing systemically with a smarter planner. Not going to stop merging though if you have strong conviction.

OmarEl-Mohandes: I really agree to your point and this why I want the planner to fail louder and clearer. This is not fixing the tool ids hallucinations.

TomSPortia (approved):

PR #135, opened 2025-02-17
Adds constraints on None values generated by the argument parser.

hannah-portia:
This toggles the `made_up` flag to trigger a clarification for None values depending on whether it's required or not. 

A local LangSmith run shows that this:
- Increases tool called by 5%
- Increases correct tool args by 3%
- Increases only_necessary_clarifications_requested by 4%

Against the most recent run

### DIFF
--- a/portia/agents/verifier_agent.py
+++ b/portia/agents/verifier_agent.py
@@ -369,6 +369,7 @@ class VerifierModel:
             for arg in tool_inputs.args
             if arg.value is None
             and arg.made_up
+            and self.agent.tool
             and not self.agent.tool.args_schema.model_fields[arg.name].is_required()
         ]
         # If a required argument value is None, we need to raise a clarification.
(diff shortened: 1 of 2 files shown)


emmaportia (dismissed):
Thanks changes look good to me, but I'd like Omar to have a look cos he spent a bunch of time on this issue

portia/agents/verifier_agent.py
```diff
@@ -362,6 +362,26 @@ def _validate_args_against_schema(self, tool_inputs: VerifiedToolInputs) -> Veri
...
+        [
+            setattr(arg, "made_up", False)
+            for arg in tool_inputs.args
+            if arg.value is None
+            and arg.made_up
+            and self.agent.tool
+            and not self.agent.tool.args_schema.model_fields[arg.name].is_required()
+        ]
+        # If a required argument value is None, we need to raise a clarification.
+        # Mark these as made up to be raised in clarifications_or_continue.
```
OmarEl-Mohandes, line 376: This case is handled already by `schema_invalid` arg that we control. I'd expect if you remove this from here, it should still give you better results (because of enforcing `made_up` to be false above).

OmarEl-Mohandes (dismissed):

TomSPortia (approved):

PR #139, opened 2025-02-18
Set up default Portia registry without Microsoft tools

robbie-portia:


portia/runner.py
```diff
@@ -69,8 +69,8 @@ class Runner:
 
     def __init__(
         self,
-        config: Config,
-        tools: ToolRegistry | list[Tool],
+        config: Config | None = None,
```
robbie-portia, line 72: I'll do a pass updating the docs to use this

portia/tool_registry.py
```diff
@@ -313,17 +314,36 @@ class PortiaToolRegistry(ToolRegistry):
...
+              the default tool set will be loaded.
 
         """
         self.api_key = config.must_get_api_key("portia_api_key")
         self.api_endpoint = config.must_get("portia_api_endpoint", str)
         self.tools = {}
-        self._load_tools()
+        self._load_tools(tool_filter or self.default_tool_filter)
+
+    @staticmethod
```
robbie-portia, line 343: Not sure if we need this method - let me know what you think

portia/runner.py
```diff
@@ -79,19 +79,26 @@ def __init__(
...
-            InMemoryToolRegistry.from_local_tools(tools) if isinstance(tools, list) else tools
-        )
+        self.config = config if config else Config.from_default()
+        logger_manager.configure_from_config(self.config)
+
+        if tools is None:
+            if self.config.portia_api_key:
+                self.tool_registry = PortiaToolRegistry.default(self.config)
+            else:
+                self.tool_registry = open_source_tool_registry
```
TomSPortia, line 89: Might be worth getting @mounir-portia view but I think we probably should alter slightly to:

```suggestion
            self.tool_registry = open_source_tool_registry
            if self.config.portia_api_key:
                self.tool_registry += PortiaToolRegistry.default(self.config)

        
            
```

TomSPortia: Might also be worth pulling this logic out into a function just to reduce the complexity in this init func?

portia/tool_registry.py
TomSPortia: I personally think we're better off pulling the logic out into a `get_default_tool_registry()` rather than polluting the ToolRegistry prototype.

portia/tool_registry.py
```diff
@@ -313,17 +314,36 @@ class PortiaToolRegistry(ToolRegistry):
...
+    def __init__(
+        self,
+        config: Config,
+        tool_filter: Callable[[Tool], bool] | None = None,
+    ) -> None:
         """Initialize the PortiaToolRegistry with the given configuration.
 
         Args:
             config (Config): The configuration containing the API key and endpoint.
+            tool_filter (Callable[[Tool], bool] | None): A filter to select a subset
```
TomSPortia, line 333: Instead of filtering at the init level what do you think about adding a method to the ToolRegistryInterface:

```python
  def filter_tools(filter_func: Callable[[Tool], bool], tool_ids: list[str]) -> ToolRegistry
```

That makes it easy to get a new subset of tools but still keep the master one around?

tests/integration/test_portia_cloud.py
```diff
@@ -76,9 +75,7 @@ def test_runner_run_query_with_cloud_and_local() -> None:
 
 def test_runner_run_query_with_oauth() -> None:
     """Test running a simple query using the Runner."""
-    config = Config.from_default()
-    tool_registry = PortiaToolRegistry(config=config)
-    runner = Runner(config=config, tools=tool_registry)
+    runner = Runner()
```
TomSPortia, line 78: 😍

portia/tool_registry.py
robbie-portia: We definitely could do. The only thing is this means that calling PortiaToolRegistry() will give back all the tools, rather than just the default set. I don't have strong feelings either way, but Emma mentioned this morning about getting the default tool registry when just init-ing the class. What do you reckon? Could get Emma's thoughts?

TomSPortia: I don't think its a big issue either way, but I think the thing its important works out of the box is the Runner() abstraction, I think its better default behaviour for PortiaToolRegistry() to give you all tools

robbie-portia: I've now updated to follow this. I have to admit, it was a bit less nice than I expected because I couldn't do a default method for filter_tools() and so I had to implement it on each class. We could change it so that registries instead implement a `remove_tool() -> ToolRegistry` and we have a default implementation of filter_tools(), but I thought it might be weird that `register_tool()` does it in-place while `remove_tool` gives you back a new registry. An alternative would be to just do the filtering inside `get_default_tool_registry` and not have this on the interface. What do you reckon?

TomSPortia: makes sense. I agree, if we're not happy in adding the interface on the base class lets just use the current approach in the get_default_tool_registry!

TomSPortia (approved):
Looks great!

portia/open_source_tools/registry.py
```diff
@@ -23,3 +31,25 @@
         FileReaderTool(),
     ],
 )
+
+
+EXCLUDED_BY_DEFAULT_TOOL_REGEXS: frozenset[str] = frozenset(
+    {
+        # Exclude Outlook by default as it clashes with Gmail
+        "portia::microsoft_outlook::*",
```
OmarEl-Mohandes, line 39: Probably best to change this to new format and wait for Hannah to push her changes (https://github.com/portiaAI/backend-holsten/pull/247/files) .. up to you.

portia/runner.py
```diff
@@ -70,8 +70,8 @@ class Runner:
 
     def __init__(
         self,
-        config: Config,
-        tools: ToolRegistry | list[Tool],
+        config: Config | None = None,
+        tools: ToolRegistry | list[Tool] | None = None,
     ) -> None:
         """Initialize storage and tools.
```
OmarEl-Mohandes, line 76: np: It might be better to document what exactly the expected defaults for config and tools that's going to be initialised from the runner since it's mostly what people interact with.

portia/runner.py
```diff
@@ -80,19 +80,23 @@ def __init__(
             tools (ToolRegistry | list[Tool]): The registry or list of tools to use.
 
         """
-        logger_manager.configure_from_config(config)
-        self.config = config
-        self.tool_registry = (
-            InMemoryToolRegistry.from_local_tools(tools) if isinstance(tools, list) else tools
-        )
+        self.config = config if config else Config.from_default()
```
OmarEl-Mohandes, line 83: Since we're talking about defaults. Do you think it's nicer to default to cloud if the portia key is present? .. if yes, then it's better to do it in the config itself.

portia/open_source_tools/registry.py
robbie-portia: Ah nice one, thanks for the heads up!

portia/runner.py
robbie-portia: Good call - this is now done

portia/runner.py
robbie-portia: This is a nice idea - mind if I do it in a follow-up PR? I think there are probably already too many changes in this one

mounir-portia (dismissed):
Looking good! We will still need to do couple tnings in docs: 1) Amend docs to reflect this wherever we currently mention or use PortiaToolRegistry, 2) Show an example usage of filter_tools e.g. to load only google tools (so like a google_tool_filter regex function + applying it in the filter_tools method for example). This latter would fit well in the tool registry section.

OmarEl-Mohandes (dismissed):

portia/open_source_tools/registry.py
```diff
@@ -23,3 +31,25 @@
...
+
+EXCLUDED_BY_DEFAULT_TOOL_REGEXS: frozenset[str] = frozenset(
+    {
+        # Exclude Outlook by default as it clashes with Gmail
+        "portia:microsoft:outlook:*",
+    },
+)
+
+
+def get_default_tool_registry(config: Config) -> ToolRegistry:
```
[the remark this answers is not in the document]
robbie-portia: Yep, definitely fair - I originally had it in there, but I ended up with a circular import from:
* `registry.py` needs to import `InMemoryToolRegistry` from `ToolRegistry`
* If we put `get_default_tool_registry` into `tool_registry.py`, than that needs to import `registry.py` to get `open_source_tool_registry`.

We could merge both files if you thought that was better and put everything in `tool_registry.py`

robbie-portia:
Updated now based on discussion in the knowledge share session

portia/open_source_tools/registry.py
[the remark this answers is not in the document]
robbie-portia: Now updated :)

TomSPortia (approved):
Lets gooo

PR #140, opened 2025-02-18
Fix LLMTool exception

robbie-portia:


OmarEl-Mohandes (approved):
Thanks for quick fix!

PR #142, opened 2025-02-18
plan builder proposal

Nathanjp91:
Added in some helper functions to help build Plans without lots of verbosity. 

```python
plan = Plan.context(
                    "Find the best offers for a flight from London to New York",
                    ["flight_search"],
                ).step(
                    "Search for flights",
                    "$flights_output",
                    tool_id="flight_search",
                ).variable(
                    "$flights_input",
                )
```


portia/plan.py
```diff
@@ -181,6 +181,89 @@ def __str__(self) -> str:
             f"steps={self.steps!r}"
         )
 
+    @classmethod
+    def context(cls, query: str, tool_ids: list[str]) -> Plan:
```
emmaportia, line 185: it feels like this would be cleaner in it's own class?

emmaportia: ack on that point. Another way to put this point is I feel without some kind of Builder interface it would be very unclear to me without reading the documentation as to what each of tehse does -- it doesn't feel like a natural builder pattern. Is there a better way you can set it up (maybe even a Builder inner class or something) to make that clear?

Nathanjp91: Swapped to use PlanBuilder class

portia/plan.py
```diff
@@ -27,6 +27,97 @@
...
+    This class provides an interface for constructing plans step by step.
+
+    """
+
+    query: str
+    tool_ids: list[str]
+    steps: list[Step]
+
+    def __init__(self, query: str) -> None:
+        """Initialize the builder with the query and tool IDs.
```
emmaportia, line 42: just the query? (Why not the toolIDs)?

Also, is it worth explaining why the system still needs the query and ToolIds given your building the plan manually?

portia/plan.py
```diff
@@ -27,6 +27,97 @@
...
+        output: str | None = None,
+        inputs: list[Variable] | None = None,
+    ) -> PlanBuilder:
+        """Add a step to the plan.
+
+        Args:
+            task (str): The task to be completed by the step.
+            tool_id (str | None): The ID of the tool used in this step, if applicable.
+            output (str | None): The unique output ID for the result of this step.
+            inputs (list[Variable] | None): The inputs to the step
```
emmaportia, line 68: explain that this creates dependencies between steps? Ie an input should always map to an output of a previous step?

portia/plan.py
```diff
@@ -27,6 +27,97 @@
...
+    steps: list[Step]
+
+    def __init__(self, query: str) -> None:
+        """Initialize the builder with the query and tool IDs.
+
+        Args:
+            query (str): The original query given by the user.
+            tool_ids (list[str]): A list of tool IDs available to the planner.
+            variables (list[Variable]): A list of variables available to the planner.
+            steps (list[Step]): A list of steps available to the planner.
```
emmaportia, line 48: mismatched args?

portia/plan.py
```diff
@@ -27,6 +27,97 @@
...
+    def build(self) -> Plan:
+        """Build the plan.
+
+        Returns:
+            Plan: The built plan.
+
+        """
+        return Plan(
+            id=PlanUUID(),
+            plan_context=PlanContext(query=self.query, tool_ids=self.tool_ids),
```
emmaportia, line 116: might be nice to have the tool_ids be the unique set of tool_ids in the steps if they haven't been provdied.

portia/plan.py
```diff
@@ -27,6 +27,97 @@
...
+
+    def build(self) -> Plan:
+        """Build the plan.
+
+        Returns:
+            Plan: The built plan.
+
+        """
+        return Plan(
+            id=PlanUUID(),
```
emmaportia, line 115: uber-nit: not needed, has a default factory

portia/plan.py
```diff
@@ -27,6 +27,97 @@
...
+
+        """
+        if len(self.steps) == 0:
+            raise ValueError("No steps in the plan")
+        if description is None:
+            description = ""
+        self.steps[-1].inputs.append(Variable(name=name, value=value, description=description))
+        return self
+
+    def build(self) -> Plan:
```
emmaportia, line 107: worht validating that all inputs have a previously defined output?

emmaportia (commented):
Thanks

portia/plan.py
Nathanjp91: The tools get dynamically added to the internal list as you add them with steps. The query is still handy because I believe it gets passed as context to the verifier agent and might be used later if we do post plan verification later.

portia/plan.py
Nathanjp91: That's basically what happens at the moment, as you add a step it gets appended to the tools list if it's not already there.

portia/plan.py
Nathanjp91: I think that should actually be done as a validator on the Plan object. I can include that as another PR but I think is seperate from the planbuilder

Nathanjp91: added that as part of a validator to Plan

portia/plan.py
Nathanjp91: variables can actually be not linked to steps but pulled from the query planner, so you can tell a plan to have an input that is not passed by a previous step. Tried to include that in the validator but it actually breaks a lot of stuff atm

portia/plan.py
```diff
@@ -145,6 +239,16 @@ class PlanContext(BaseModel):
     query: str = Field(description="The original query given by the user.")
     tool_ids: list[str] = Field(description="The list of tools IDs available to the planner.")
 
+    @field_serializer("tool_ids")
+    def serialize_tool_ids(self, tool_ids: list[str]) -> list[str]:
```
Nathanjp91, line 243: useful for comparison and equality checking

portia/plan.py
```diff
@@ -22,11 +22,105 @@
...
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, ConfigDict, Field, field_serializer, model_validator
 
 from portia.prefixed_uuid import PlanUUID
 
 
+class PlanBuilder:
+    """A builder for creating plans.
+
+    This class provides an interface for constructing plans step by step. Requires a steps to be
```
emmaportia, line 33: s/steps/step

portia/plan.py
```diff
@@ -22,11 +22,105 @@
...
+    added to the plan before building it and mapping inputs and outputs to the correct steps.
```
emmaportia, line 34: i don't think 'mapping inputs and outputs to the correct steps.' would necessarily make sense without context?

portia/plan.py
emmaportia: (Update the comment tho cos you aren't initializing it with tool IDs)

portia/plan.py
```diff
@@ -22,11 +22,105 @@
...
+        self.steps.append(Step(task=task, output=output, inputs=inputs, tool_id=tool_id))
+        return self
+
+    def input(
+        self,
+        name: str,
+        value: Any | None = None,  # noqa: ANN401
+        description: str | None = None,
+    ) -> PlanBuilder:
+        """Add an input variable to the last step in the plan.
```
emmaportia, line 92: to the _current_ last step in the plan builder?

emmaportia: you defo think it's worth this complexity? should we keep it out for now and if users want it we can add it back in? i just feel like the manual input concept is unlikely to be used.

emmaportia: would it better to have a step index for which step to add it to, maybe default to the last step?

emmaportia (commented):
Thanks and apologies for being slow on this

portia/plan.py
Nathanjp91: added a step index. But we can't remove this because there needs to be some way to map inputs, unless we take a naive approach of just assuming every new step that gets added inherits all outputs of previous steps as an input. That might be doable but feels like a lot of interplay with execution agent and runner and context windows, it would severely limit how big any manually build plans could be because there'd not be any way to limit what gets fed to the execution agent. 

We could have other ways to codify that though with some default assumptions, like assuming if nothing is given that it takes the previous step as input, and maybe like 

```python
plan = PlanBuilder("example plan")
    .step(...)
    .step(...)
        .relies_on([0]) # step 0
    .step(...) # assumes input from previous step
    .step(...)
        .relies_on([0, 2]) # steps 0 and 2
```

I'm not convinced this is easier to understand than input though

portia/plan.py
```diff
@@ -22,11 +22,111 @@
...
+        description: str | None = None,
+        step_index: int | None = None,
+    ) -> PlanBuilder:
+        """Add an input variable to the chosen step in the plan (default is the last step).
+
+        Args:
+            name (str): The name of the input.
+            value (Any | None): The value of the input.
+            description (str | None): The description of the input.
+            step_index (int | None): The index of the step to add the input to.
```
emmaportia, line 99: say if 'None' defaults to the last step?

portia/plan.py
emmaportia: i think you could still say something along the lines of 'Can be used to create dependencies between steps'?

portia/plan.py
```diff
@@ -22,11 +22,111 @@
...
 
 
+class PlanBuilder:
+    """A builder for creating plans.
+
+    This class provides an interface for constructing plans step by step. Requires a step to be
+    added to the plan before building it.
+
+    Example:
+    ```python
```
emmaportia, line 37: i *think* this syntax breaks docusaurus from memory -- can you check if the SDK docs still build?

emmaportia (commented):
Thanks. Mostly nits, just the comment on docusaurus which i wanted to double check.

emmaportia (approved):
Thanks!

PR #146, opened 2025-02-19
Verifier clarification prompt improvements

hannah-portia:
Modifies the verifier agent prompt to address common error patterns from evals relating to clarifications.

This includes:
- Not providing a "default" value for required arguments.
- Avoiding type swaps between the argument and verifier parsers.
- Reducing the rate of wrong arugments marked as made up by referring to the goal as well as the context.

LangSmith evals show increase to 88% correct tool arguments mostly from an increase to the volume of expected clarifications raised (to 98.8%)
[LangSmith eval](https://smith.langchain.com/o/e2eb5117-7ff4-446a-b9ac-b768f33886e1/datasets/501d558e-5303-4393-aca4-547cd8c2325c/compare?selectedSessions=258f3118-3746-495e-8ce9-575c92b6fd12&baseline=undefined)

[Comparison to a baseline run](https://smith.langchain.com/o/e2eb5117-7ff4-446a-b9ac-b768f33886e1/datasets/501d558e-5303-4393-aca4-547cd8c2325c/compare?selectedSessions=258f3118-3746-495e-8ce9-575c92b6fd12&baseline=undefined) of latest Portia shows a ratio for correct tool args is 19:6 (for every 1 regression, we see 3.2 improvements).
For expected clarifications raised, the ratio is 21:0, which is great to see no regressions.


portia/agents/verifier_agent.py
```diff
@@ -272,18 +277,22 @@ class VerifierModel:
...
+                "Prefer the type of the argument provided in the list of arguments provided to "
+                "label at the end of the response. These have been through an argument parser and "
+                "are more likely to be correct.\n"
                 "  made_up: bool  # if the value is made_up based on the given rules.\n\n"
                 "class VerifiedToolInputs:\n"
                 "  args: List[VerifiedToolArgument]  # List of tool arguments.\n\n"
                 "Please ensure the output matches the VerifiedToolInputs schema.",
             ),
             HumanMessagePromptTemplate.from_template(
+                "You will need to achieve the following goal: {task}\n"
```
OmarEl-Mohandes, line 290: I've read on couple of forms before that tasks would better placed at the end of the prompt for better results. Feel free to play with it if you have time, otherwise I'll cover it in my context management experiments.

OmarEl-Mohandes (approved):
Nice results! .. Thanks!

portia/agents/verifier_agent.py
hannah-portia: In my error analysis, I noticed it was not using information from the goal when deciding whether something is made_up. I tried both pulling it forward and writing in the prompt explicitly to look at the goal when considering explanations.

We should have a revisit to this at some point

PR #148, opened 2025-02-19
Update README.md

mounir-portia:
fixed parameter llm_model (should be llm_model_name now) in the second code block.


robbie-portia (approved):
Good catch!

PR #151, opened 2025-02-20
Allow running different LLMs in different areas (and change defaults)

robbie-portia:
This PR:
* Allows specifying different LLMs for the different areas 
* Enabled the use of o3-mini. This meant removing temperature as a potential configuration option (as you can't set temperature with o3-mini)
* Enabled o3 mini as the default planner module, and 4o as the default agent module (as these were where I got the best results on the evals)
* Removed the LLM seed option - otherwise it felt like we had too many config options and I couldn't imagine many people used it.
* Also includes a few small prompt improvements to the verifier agent as I was hill climbing on the agent eval (sorry, I probably should have split these out!)


portia/agents/verifier_agent.py
```diff
@@ -272,7 +272,8 @@ class VerifierModel:
                 "\n- If an argument is marked as invalid it is likely wrong."
                 "\n- We really care if the value of an argument is not in the context, a handled "
                 "clarification or goal at all (then made_up should be TRUE), but it is ok if "
-                "it is there but in a different format (then made_up should be FALSE). "
+                "it is there but in a different format, or if it can be reasonably derived from the"
```
robbie-portia, line 275: The changes to this prompt were small ones I made as part of improving the agent evals. I'm happy to split them out into a separate PR if that's easiest.

portia/cli.py
```diff
@@ -284,32 +284,10 @@ def _get_config(
         load_dotenv(override=True)
     config = Config.from_default(**kwargs)
 
-    keys = [
```
robbie-portia, line 287: I thought this could be removed - we can leave the config handling to the config class.

portia/config.py
```diff
@@ -102,7 +86,7 @@ class LLMModel(Enum):
         GPT_3_5_TURBO: GPT-3.5 Turbo model by OpenAI.
         CLAUDE_3_5_SONNET: Claude 3.5 Sonnet model by Anthropic.
         CLAUDE_3_5_HAIKU: Claude 3.5 Haiku model by Anthropic.
-        CLAUDE_3_OPUS_LATEST: Claude 3.0 Opus latest model by Anthropic.
+        CLAUDE_3_OPUS: Claude 3.0 Opus model by Anthropic.
```
robbie-portia, line 89: Minor renaming to be consistent with the others

portia/config.py
```diff
@@ -195,27 +181,6 @@ class LogLevel(Enum):
     CRITICAL = "CRITICAL"
 
 
-def is_greater_than_zero(value: int) -> int:
```
robbie-portia, line 198: This wasn't needed after I removed temperature (and the 100% test coverage check complained)

portia/config.py
```diff
@@ -528,13 +539,62 @@ def default_config(**kwargs) -> Config:  # noqa: ANN003
         Config: The default config
 
     """
+    # Precedence for LLM config is:
```
robbie-portia, line 542: I definitely don't love the complexity of this - let me know if you can think of a better way

hannah-portia (approved):

hannah-portia:
Sorry, accidentally pushed approve on the wrong PR

portia/agents/verifier_agent.py
hannah-portia: I think @OmarEl-Mohandes has recently made some changes to the prompt you might want to sync in

portia/config.py
hannah-portia: Should the MISTRAL latest below also be renamed then?

portia/config.py
```diff
@@ -269,10 +234,10 @@ class Config(BaseModel):
         default_log_level: The default log level (e.g., DEBUG, INFO).
         default_log_sink: The default destination for logs (e.g., sys.stdout).
         json_log_serialize: Whether to serialize logs in JSON format.
-        llm_provider: The LLM provider (e.g., OpenAI, Anthropic).
-        llm_model_name: The model to use for LLM tasks.
-        llm_model_temperature: The temperature for LLM generation.
-        llm_model_seed: The seed for LLM generation.
+        planner_llm_model_name: The specific LLM model used for the planner agent.
```
hannah-portia, line 237: What are people's thoughts on starting to add some hierarchy here?

Once we support more LL providers or more agent types, this could get quite large

portia/config.py
```diff
@@ -349,40 +314,31 @@ def parse_default_log_level(cls, value: str | LogLevel) -> LogLevel:
         default=False,
         description="Whether to serialize logs to JSON",
     )
-
-    # LLM Options
-    llm_provider: LLMProvider = Field(
-        default=LLMProvider.OPENAI,
-        description="Which LLM Provider to use.",
+    planner_llm_model_name: LLMModel = Field(
+        description="Which LLM Model to use for the planner.",
```
hannah-portia, line 318: Do we want to put the agent terminology in the descriptions (and or variable name)?

portia/config.py
```diff
@@ -520,6 +480,57 @@ def must_get(self, name: str, expected_type: type[T]) -> T:
...
+        match model_name.provider():
+            case LLMProvider.OPENAI:
+                return self.openai_api_key
+            case LLMProvider.ANTHROPIC:
+                return self.anthropic_api_key
+            case LLMProvider.MISTRALAI:
+                return self.mistralai_api_key
+
+
+PLANNER_DEFAULT_MODELS = {
```
hannah-portia, line 499: Can we have a shared default and then just override where necessary? It's all the same except for the planner openAI model

portia/config.py
```diff
@@ -528,13 +539,62 @@ def default_config(**kwargs) -> Config:  # noqa: ANN003
         Config: The default config
 
     """
+    # Precedence for LLM config is:
+    # 1. Specific LLM model name - e.g. planner_llm_model_name="gpt-4o"
+    # 2. General LLM model name - e.g. llm_model_name="gpt-4o"
```
hannah-portia, line 544: Is the difference b/w 1 and 2 that someone specifies a model per agent?

tests/integration/test_llm_wrapper.py
```diff
@@ -4,36 +4,25 @@
...
-    (
-        LLMProvider.MISTRALAI,
-        LLMModel.MISTRAL_LARGE_LATEST,
-    ),
-    (
-        LLMProvider.ANTHROPIC,
-        LLMModel.CLAUDE_3_OPUS_LATEST,
-    ),
+MODELS = [
+    LLMModel.GPT_4_O_MINI,
```
hannah-portia, line 12: I saw above we're using 4O as default and we're not using mini/small models for the other providers here. Should we swap?

portia/config.py
robbie-portia: Yeah, I thought I'd do that in a follow-up PR - the mistral model is used in lots of tests and this PR was already getting too big.

portia/agents/verifier_agent.py
robbie-portia: I had a quick look and I think it looked like they play ok together. What do you think @OmarEl-Mohandes ?

portia/config.py
robbie-portia: Good call. I think it's ok for the argument to be `planning` and not mention agent as it saves characters (same as it's just `execution` below) but I've updated the description.

portia/config.py
robbie-portia: Good call - done

portia/config.py
robbie-portia: Yep, exactly 👍  So if you did e.g. `--planning_llm_model_name="o3-mini" --llm_model_name="gpt-4o"` then the planner agent would get `o3-mini` and the others would get `gpt-4o`

tests/integration/test_llm_wrapper.py
robbie-portia: Done 👍

portia/config.py
robbie-portia: Yeah, I'm interested in this. I think this is something we should do when we delete `LLMWrapper` and generally clean up this code, but happy to try and do something for now if people prefer

portia/config.py
```diff
@@ -520,6 +480,45 @@ def must_get(self, name: str, expected_type: type[T]) -> T:
...
+
+def llm_provider_default_from_api_keys() -> LLMProvider:
+    """Get the default LLM provider from the API keys."""
+    if os.getenv("OPENAI_API_KEY"):
+        return LLMProvider.OPENAI
+    if os.getenv("ANTHROPIC_API_KEY"):
+        return LLMProvider.ANTHROPIC
+    if os.getenv("MISTRAL_API_KEY"):
+        return LLMProvider.MISTRALAI
+    return LLMProvider.OPENAI
```
Nathanjp91, line 520: if openai key hasn't been provided by this point, I feel like it should raise instead

portia/agents/verifier_agent.py
```diff
@@ -572,7 +579,10 @@ def execute_sync(self) -> Output:
         context = self.get_system_context()
         execution_context = get_execution_context()
         execution_context.workflow_run_context = context
-        llm = LLMWrapper(self.config).to_langchain()
+        llm = LLMWrapper(
+            self.config.execution_llm_model_name,
```
Nathanjp91, line 583: I'm not a huge fan of this but I also don't really like llmwrapper generally, so hard to disentangle my dislikes. To me I think our process should be like 
```python
config = Config(execution_llm_model_name='openai/o3-mini')
assert isinstance(config.execution_llm_model, BaseChatModel)
```
config takes a string, but actually builds the llm model in the config init, then everywhere else instead of repeating all this llmwrapper and to_langchain calls, we just pull the config model. Ideally this would be like, a litellm model from str instead, but for now we're tied to langchain

Nathanjp91: and llmwrapper basically only then gets called in the model init code of config

Nathanjp91: if we need them built on the fly for whatever reason, we can make them 
@property 
on the config class

this then cleans up a lot of code in the interim

```python
llm = self.config.execution_model

# where 
class Config:
    ...
    @property
    def execution_model():
        return LLMWrapper(self.execution_llm_model_name...)
```

portia/config.py
```diff
@@ -291,15 +256,15 @@ class Config(BaseModel):
     )
 
     # LLM API Keys
-    openai_api_key: SecretStr | None = Field(
+    openai_api_key: SecretStr = Field(
```
Nathanjp91, line 259: these are getting a bit complicated now. I think we should move to an api_keys: dict[str, SecretStr] type manager, but maybe not in this PR.

portia/config.py
```diff
@@ -349,40 +314,31 @@ def parse_default_log_level(cls, value: str | LogLevel) -> LogLevel:
         default=False,
         description="Whether to serialize logs to JSON",
     )
-
-    # LLM Options
-    llm_provider: LLMProvider = Field(
-        default=LLMProvider.OPENAI,
-        description="Which LLM Provider to use.",
+    planning_llm_model_name: LLMModel = Field(
```
Nathanjp91, line 317: kind of the same comment as above. Then we can do defaults as well with the properties approach

```python
models: dict[str, str] = ...
@property
def execution_model():
    if 'execution_model' in self.models:
        return LLMWrapper(self.models[execution_model]
    return LLMWrapper(self.models['default_model']
```


This is then extensible, perhaps as we add more agents to the system, you can dynamically plug them in as that grows per our naming, ie as in when we add a new agent , introspection or whatever, we just have to give it a name wherever we're calling that agent, and config can already handle it without updating, and if it can't find one it just uses a default

portia/config.py
Nathanjp91: see my comments about moving to models dictionary above

Nathanjp91 (changes requested):
Can discuss in person tomorrow. I think there's lots we can improve here but like the freedom that the model changes will allow us.

portia/config.py
Nathanjp91: actually to make it more extensible it would probably just be 

```python
class Config:
    def model(self, name: str) -> BaseChatModel:
        if 'execution' in self.models:
            return ...
        return ...
```

portia/agents/verifier_agent.py
robbie-portia: Discussed - let's go with: https://github.com/portiaAI/portia-sdk-python/pull/151#discussion_r1972590032

portia/llm_wrapper.py
```diff
@@ -151,30 +153,33 @@ def to_langchain(self) -> BaseChatModel:
...
-                    temperature=self.model_temperature,
+                    name=self.model_name.value,
+                    model=self.model_name.value,
                     seed=self.model_seed,
-                    api_key=self.config.openai_api_key,
+                    api_key=self.api_key,
                     max_retries=3,
+                    # Unfortunately you get errors from o3 mini with Langchain unless you set
+                    # temperature to 1. See https://github.com/ai-christianson/RA.Aid/issues/70
+                    temperature=1 if self.model_name == LLMModel.O3_MINI else 0,
```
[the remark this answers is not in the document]
robbie-portia: That's fair, though I think with the way things are currently setup it'd be v difficult to allow setting a temperature for each of the different LLMs. I raised https://linear.app/portialabs/issue/POR-886/remove-llm-wrapper-tidy-config to allow this as a follow-up and put it on next cycle. What do you think?

portia/config.py
robbie-portia: Done

portia/agents/verifier_agent.py
robbie-portia: I've now mostly done this. The part that didn't quite work was to have `Config` return an `LLMWrapper` - that introduced a circular dependency because then `Config` relies on `LLMWrapper` and `LLMWrapper` already relies on `Config` for e.g. switching on the different LLM providers. I changed things so `LLMWrapper` now has a `for_usage` method.

I've also raised https://linear.app/portialabs/issue/POR-886/remove-llm-wrapper-tidy-config which includes remove LLMWrapper.

portia/config.py
robbie-portia: Yeah, I'm keen for this PR to not grow further, but included this in https://linear.app/portialabs/issue/POR-886/remove-llm-wrapper-tidy-config

portia/execution_agents/one_shot_agent.py
```diff
@@ -177,7 +178,7 @@ def execute_sync(self) -> Output:
 
         context = self.get_system_context()
 
-        llm = LLMWrapper(self.config).to_langchain()
+        llm = LLMWrapper.for_usage(LLMUsage.EXECUTION, self.config).to_langchain()
```
Nathanjp91, line 181: So here, I'd prefer the agent get a new name field that was static

```
self.name = "EXECUTION" 
...
llm = self.config.model(self.name).to_langchain()
```
New agents just need to be named and the config model dict will pick it up automatically, no imports required.

portia/open_source_tools/llm_tool.py
```diff
@@ -56,8 +57,7 @@ class LLMTool(Tool[str]):
 
     def run(self, ctx: ToolRunContext, task: str) -> str:
         """Run the LLMTool."""
-        llm_wrapper = LLMWrapper(ctx.config)
-        llm = llm_wrapper.to_langchain()
+        llm = LLMWrapper.for_usage(LLMUsage.LLM_TOOL, ctx.config).to_langchain()
```
Nathanjp91, line 60: This then gives freedom to define tools that actually pull llms from the config that aren't registered. You could feasibly add 
```
config.models = {... "image_processing": LLMModel(...)}
```
And then in another part of the system, connect tools that use this

```
class LLMImageProcessor():
    ...
    def run(...):
        llm = ctx.config.model("image_processing").to_langchain()
```
without having to formally codifying the relationship, which requires a lot of abstraction class updates.

portia/config.py
```diff
@@ -292,19 +298,37 @@ class Config(BaseModel):
...
+        default=LLMProvider.OPENAI,
+        description="Which LLM Provider to use.",
+    )
+
+    models: dict[LLMUsage, LLMModel] = Field(
+        default={},
+        description="A dictionary of configured LLM models for each usage.",
+    )
+
+    def model(self, usage: LLMUsage) -> LLMModel:
```
Nathanjp91, line 324: I would just simplify this by codifying the defaults as part of config validation

```python
class Config():
    ...
    @model_validator(mode='after')
    def validate():
        ...
        default_planner = PLANNER_DEFAULT_MODELS[self.llm_provider]
        self.models = {
            'planning_default': default_planner,
            'default': default,
            **self.models
            }
            return self
  ```
  
  Then models just becomes
  ```python
  def models(self, name: str) -> LLMModel:
      if name == "PLANNING":
          return self.models.get("PLANNING", self.models.get("PLANNING_DEFAULT"))
      return self.models.get(name, self.models.get("DEFAULT"))
```

Nathanjp91 (changes requested):
Happy to debate over these, but generally I think we should be moving away from very strictrly codifying llm and agent relationships and away from abstractions, not more. The typing information is a minor loss there, but the flexibility gained for users is worth that imo.

portia/execution_agents/one_shot_agent.py
robbie-portia: Now done 👍  The one bit that's different is that I still use the `LLMWrapper.for_usage()` to avoid the circular import issue (`LLMWrapper` still needs config, so we can't have `LLMWrapper` inside config).

portia/open_source_tools/llm_tool.py
robbie-portia: This is v nice 🚀

portia/config.py
robbie-portia: Done

Nathanjp91 (approved):
Like it: Thanks for patience with the feedback!

PR #153, opened 2025-02-21
Runs evals on portia-sdk-python PRs

hannah-portia:
Uses the GitHub Action in the evals repo to kick-off evals prior as a blocker to PRs.

Note: This will need to wait for the evals Action to be submitted in the evals repo.

Towards POR-768

### DIFF
--- a/.github/workflows/evals.yml
+++ b/.github/workflows/evals.yml
@@ -2,6 +2,8 @@ name: Run Evals for portia-sdk-python
 
 on:
   pull_request_target:
+    branches:
+      - main
     types:
       - opened
       - synchronize
--- a/.github/workflows/evals.yml
+++ b/.github/workflows/evals.yml
@@ -1,13 +1,34 @@
 name: Run Evals for portia-sdk-python
 
 on:
-  workflow_dispatch:
-  pull_request:
+  pull_request_target:
     branches:
       - main
+    types:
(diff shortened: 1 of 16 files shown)


.github/workflows/evals.yml
```diff
@@ -0,0 +1,19 @@
+name: Run Evals for portia-sdk-python
+
+on:
+  workflow_dispatch:
+  pull_request:
```
TomSPortia, line 5: I think here we might need to do `pull_request_target` like we do in the tests one with the `safe_to_test` library. Otherwise the secrets wont exist

hannah-portia: Re-checking for the same label makes sense right? It doesn't make sense to have a separate label for evals if we need both to pass and the label is more to say that it's safe IIUC.

I think it makes sense to restrict it to main just because it's going to cost more than other actions.

hannah-portia: Are only we able to add this label?

TomSPortia: yep only maintainers and yep same label 👍

hannah-portia: Awesome, just popped the changes up

TomSPortia (approved):

TomSPortia (approved):

PR #157, opened 2025-02-24
Add clarification handling to `execute_workflow`

robbie-portia:
Raising this after a conversation I had with Emma / Mounir last week about the fact that ideally users wouldn't have to write the `while workflow.state == NEED_CLARIFICATION` loop. This PR brings that inside `execute_workflow`, and adds a `ClarificationHandler` class that allows the user to specify how clarifications should be handled. This also allows us to provide default clarification handling.

I'm unsure of this, so raising this PR to start discussion, get feedback and also see if there are better ways. The pros of this approach are that we remove the burden of writing the clarification handling loop from the user and also that we can provide default clarification handling. The cons are that it makes things a bit more opaque, potentially gives the user less control over the clarification handling flow and the user now has to understand the `ClarificationHandler` class if they want to handle clarifications.

What do you think? Any suggestions on how to do this in a better way?


TomSPortia:
I like the direction of this but I think we should generalise the approach and combine it with: https://linear.app/portialabs/issue/POR-712/runner-hooks-per-step and have the idea of a set of hooks:

From the top of my head we can have:

- OnClarification
- OnStepComplete
- OnStepFailed

Each of which is given the whole workflow and can update the state?

On the substance of the PR, I'm not a fan of the default implementation relying on Click which is only really used for the CLI and doesn't make sense in a server context etc. I actually think we should not ship a default implementation. We can provide a log based one in the SDK but I think we should force the user to think about how they'll handle clarifications before they run a workflow. 

I also think we should follow the interface approach of the rest of the PR for the handlers.

portia/clarification_handler.py
```diff
@@ -0,0 +1,147 @@
...
+                    cast(ValueConfirmationClarification, clarification),
+                )
+            case ClarificationCategory.CUSTOM:
+                return self.handle_custom_clarification(
+                    runner,
+                    workflow,
+                    cast(CustomClarification, clarification),
+                )
+            case ClarificationCategory.ARGUMENT:
+                raise NotImplementedError("Argument clarification not implemented")
```
robbie-portia, line 75: I wasn't sure what to do in this case - when would we expect the user to handle this clarification?

robbie-portia:
Yeah, I agree this potentially fits well with the hooks approach 👍 And that's fair on `click` - we could have `ClarificationHandler` be an ABC, but then offer a `CLIClarificationHandler` which people can use when running locally to get themselves setup quickly.

emmaportia:
> I like the direction of this but I think we should generalise the approach and combine it with: https://linear.app/portialabs/issue/POR-712/runner-hooks-per-step and have the idea of a set of hooks:
> 
> From the top of my head (haven't thought deeply) we can have:
> 
> * OnClarification
> * OnStepComplete
> * OnStepFailed
> * OnWorkflowComplete
> 
> Each of which is given the whole workflow and can update the state?
> 
> On the substance of the PR, I'm not a fan of the default implementation relying on Click which is only really used for the CLI and doesn't make sense in a server context etc. I actually think we should not ship a default implementation. We can provide a log based one in the SDK but I think we should force the user to think about how they'll handle clarifications before they run a workflow.
> 
> I also think we should follow the interface approach of the rest of the PR for the handlers.

I agree with Tom generally here, we had something akin to this in [low_code_agent](https://github.com/portiaAI/low_code_agent/blob/main/portia/cli_runner.py) with one caveat that i think we need to make sure it's simple... I worry sometimes when we keep adding abstactions particularly ones with hooks and callbacks, that it might be hard to understand for a first time user and there's already some pretty complicated control flow stuff that goes on under the hood. So generally in alignment on direction, but would be dependent on the complexity of the implementation.

portia/clarification_handler.py
robbie-portia: Looking at the code, we could potentially make it abstract or even remove. What do people think?

robbie-portia:
Also, on a completely separate note, how do people usually handle things like the CLIClarificationHandler with the 100% test coverage check? Would people write a test that mocks out click, or bypass the check somehow (I suspect there's a way to do this as I doubt we're checking every line of e.g. cli.py, but maybe I'm wrong)?

portia/clarification_handler.py
TomSPortia: Yeah I think it was conceived as an abstract class, people shouldn't raise an argument clarification but a input or multi choice etc

TomSPortia: I think we should remove it as a category

portia/clarification_handler.py
```diff
@@ -0,0 +1,200 @@
...
+    def handle_custom_clarification(
+        self,
+        runner: Runner,
+        workflow: Workflow,
+        clarification: CustomClarification,
+    ) -> Workflow:
+        """Handle a custom clarification."""
+
+
+class CLIClarificationHandler(ClarificationHandler):
```
TomSPortia, line 124: I think it makes more sense for this to live in the CLI.py file?

portia/runner.py
```diff
@@ -250,11 +261,29 @@ def execute_workflow(
...
+        """Execute a workflow and handle any clarifications that are raised."""
         # if there is execution context set, make sure we update the workflow before running
         workflow.execution_context = get_execution_context()
-        return self._execute_workflow(plan, workflow)
+        workflow = self._execute_workflow(plan, workflow)
+
+        while workflow.state == WorkflowState.NEED_CLARIFICATION:
+            # If clarifications are needed, resolve them before resuming the workflow
+            for clarification in workflow.get_outstanding_clarifications():
+                workflow = self.clarification_handler.handle(self, workflow, clarification)
```
TomSPortia, line 281: Worth checking after all the handlers have run that the clarifications have actually been resolved? If not throw an error?

portia/runner.py
```diff
@@ -250,11 +261,29 @@ def execute_workflow(
...
+
+            # Once clarifications are resolved, resume the workflow
+            workflow = self.execute_workflow_and_handle_clarifications(plan, workflow)
```
TomSPortia, line 284: I don't mind the recursion but I wonder if it'd be cleaner to change the condition on line #278 to be while workflow.state not in [COMPLETE, FAILED] and then just loop in a single function?

portia/clarification_handler.py
```diff
@@ -0,0 +1,200 @@
...
+    def handle(self, runner: Runner, workflow: Workflow, clarification: Clarification) -> Workflow:
+        """Handle a clarification by routing it to the appropriate handler.
+
+        Args:
+            runner: The runner that is running the workflow
+            workflow: The workflow that the clarification was raised on
+            clarification: The clarification object to handle
+
+        """
+        match clarification.category:
```
TomSPortia, line 44: I think if you match on the type here rather than the category you'll be able to lose the `cast` call

portia/clarification_handler.py
```diff
@@ -0,0 +1,200 @@
...
+from portia.workflow import Workflow, WorkflowState
+
+if TYPE_CHECKING:
+    from portia.runner import Runner
+
+
+class ClarificationHandler:
+    """Handles clarifications that arise during the execution of a workflow."""
+
+    def handle(self, runner: Runner, workflow: Workflow, clarification: Clarification) -> Workflow:
```
TomSPortia, line 35: I can see why we did it (wait_for_ready) but I don't really like the runner being passed to the handler. In my head these handlers/hooks should be able to update the workflow state, but actually persisting that state should be a runner task. 

I don't have an obviously better option but just concerned that this feels like it tightly couples abstractions in a way I've tried to avoid elsewhere.

portia/runner.py
```diff
@@ -72,6 +73,7 @@ def __init__(
         self,
         config: Config | None = None,
         tools: ToolRegistry | list[Tool] | None = None,
+        clarification_handler: ClarificationHandler | None = None,
```
TomSPortia, line 76: What do you think about introducing a type:

```
class ExecutionHooks():
  clarification_handler: ClarificationHandler | None
```

and having that as whats passed in here. It leaves the door open for us in the future to have other hooks we add.

portia/clarification_handler.py
TomSPortia: I think it might be worth two minutes just to see if `wait_for_ready` could be pulled out into a function rather than on the runner?

Given this approach people are less likely to directly call it, and I think if we move the storage calls outside of the function the core of the code doesn't need runner primitives.

TomSPortia: Or if thats not easy instead of passing the whole runner perhaps we can define an interface thats a subset of the runner with only what the handler needs.

portia/runner.py
```diff
@@ -81,10 +83,16 @@ def __init__(
...
                 from Portia cloud if a Portia API key is set.
+            clarification_handler (ClarificationHandler | None): The clarification handler
+                to use when clarifications are raised. If not provided, the default
+                clarification handler will be used.
 
         """
         self.config = config if config else Config.from_default()
         logger_manager.configure_from_config(self.config)
+        self.clarification_handler = (
+            clarification_handler if clarification_handler else ClarificationHandler()
```
TomSPortia, line 94: The default here is the abstract class? I think if we're going to provide a default it needs to work, otherwise we should make it required.

TomSPortia (commented):
I'm just a bit worried about the clarification_handler <-> runner dependency linking, otherwise I think this is heading in a good direction.

portia/clarification_handler.py
robbie-portia: Done 👍

portia/runner.py
robbie-portia: This now calls self.wait_for_ready(workflow) which raises if they haven't been resolved.

portia/runner.py
robbie-portia: Updated 👍

portia/clarification_handler.py
robbie-portia: Done

portia/runner.py
robbie-portia: Done

portia/runner.py
robbie-portia: I've made a few changes now relevant to this:
* We allow not passing in a clarification handler - if you don't, then the flow works as it did previously.
* I've changed the handling in `Runner` so that we now pass `resolve()` and `error()` methods to the clarification handler, and changed it to be the runner's job to wait for a resolution. I think that works well for resolving (or erroring) clarifications async - you can just call that `resolve()` or `error()` method whenever you like.

What do you think?

portia/clarification.py
```diff
@@ -79,7 +79,7 @@ class Clarification(BaseModel, Generic[SERIALIZABLE_TYPE_VAR]):
     )
 
 
-class ArgumentClarification(Clarification[SERIALIZABLE_TYPE_VAR]):
+class ArgumentClarification(Clarification[SERIALIZABLE_TYPE_VAR], ABC):
```
emmaportia, line 82: why do we still need this?

portia/clarification_handler.py
```diff
@@ -0,0 +1,122 @@
...
+        resolve: Callable[[Clarification, object], None],
+        error: Callable[[Clarification, object], None],
+    ) -> None:
+        """Handle a clarification by routing it to the appropriate handler.
+
+        Args:
+            clarification: The clarification object to handle
+            resolve: Callback function to resolve the clarification. This can either be called
+                synchronously in this function or called async after returning from this function.
+            error: Callback function to mark that handlign a clarification has failed. This can
```
emmaportia, line 38: nit: s/handlign/handling

portia/clarification_handler.py
```diff
@@ -0,0 +1,122 @@
...
+        self,
+        clarification: Clarification,
+        resolve: Callable[[Clarification, object], None],
+        error: Callable[[Clarification, object], None],
+    ) -> None:
+        """Handle a clarification by routing it to the appropriate handler.
+
+        Args:
+            clarification: The clarification object to handle
+            resolve: Callback function to resolve the clarification. This can either be called
```
emmaportia, line 36: "Callback function that should be invoked once the clarification has been handled, prompting the plan run to resume"?

I would probably want to call the resolve and error functions something more callbacky, e.g 'on_resolution', or 'on_error'?

portia/clarification_handler.py
```diff
@@ -0,0 +1,122 @@
...
+                synchronously in this function or called async after returning from this function.
+            error: Callback function to mark that handlign a clarification has failed. This can
+                either be called synchronously in this function or called async after returning
+                from this function.
+
+        """
+        match clarification:
+            case ActionClarification():
+                return self.handle_action_clarification(
```
emmaportia, line 45: optional nit, you could just set the function and then invoke it with the clarification, resolve, error args rather than repeating it for each type, e.g:

```
match clarification:
    case ActionClarification():
        f = self.handle_action_clarification
    ...

    f(clarification, resolve, error)
```

portia/runner.py
```diff
@@ -58,10 +58,19 @@
 
 if TYPE_CHECKING:
     from portia.agents.base_agent import BaseAgent
+    from portia.clarification_handler import ClarificationHandler
     from portia.planners.planner import Planner
     from portia.tool import Tool
 
 
+class ExecutionHooks:
+    """Hooks that can be used to add modify or add extra functionality to the execution of a workflow."""  # noqa: E501
```
emmaportia, line 67: Worth adding a bit more sub detail to explain they can be used to handle clarifications since that's the only use currently?

emmaportia (commented):
thanks!

portia/clarification.py
robbie-portia: I don't think we really do - 3 of the other clarifications were inheriting from it, but I've now remove it

portia/clarification_handler.py
robbie-portia: Unfortunately the types complain if we do this because the clarification argument in each case are the different subclasses of `Clarification` (e.g. `handle_action_clarification` takes an `ActionClarification`, `handle_input_clarification` takes an `InputClarification` etc)

portia/runner.py
robbie-portia: I've added a bit more - let me know if there are any more specifics you think it's worth mentioning in there.

portia/clarification_handler.py
robbie-portia: Done

emmaportia (approved):

PR #162, opened 2025-02-25
Make non-alpha release

emmaportia:
What I'm hoping will happen here:
- It will automatically pick up the new tag, and publish this to pypi with the new release number

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "portia-sdk-python"
-version = "0.0.1-alpha.58"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "portia-sdk-python"
-version = "0.0.1-alpha.58"
+version = "0.0.1.58"
```
TomSPortia, line 3: I don't think this works with semantic versioning. I think what we want is `0.0.1` which would be the first minor release or more likely `1.0.0` for the first major release?

PR #163, opened 2025-02-25
Reduces Search Tool results emitted

hannah-portia:
Having a large number of results is resulting in poor performance in the parser in the VerifierModel for downstream tools utilising the search results (usually the LLMTool)


portia/open_source_tools/search_tool.py
```diff
@@ -57,5 +57,8 @@ def run(self, _: ToolRunContext, search_query: str) -> str:
         response.raise_for_status()
         json_response = response.json()
         if "answer" in json_response:
-            return json.dumps(json_response, indent=2)
+            results = json_response["results"]
+            if len(results) > MAX_RESULTS:
+                return results[:MAX_RESULTS]
```
OmarEl-Mohandes, line 62: You don't need the if condition it will just take the minimum between `MAX_RESULTS` and len(results) when you call `results[:MAX_RESULTS]`

OmarEl-Mohandes (approved):

OmarEl-Mohandes (commented):
If you need these changes to be available for Mounir..etc, you need to bump the version.

hannah-portia:
It's fine for this to come out in the next release

PR #168, opened 2025-02-26
Fixes eval action in portia SDK

hannah-portia:
It's not possible to execute an action from a private repo in a public one. To avoid needing to migrate to Cloud Enterprise to get access to internal repo (which can be run from a public repo), this action now checks out the evals repo and duplicates the same action.

This requires the same safe_to_test label to be set before running as the testing framework. 
Tested manually via workflow dispatch (now removed): https://github.com/portiaAI/portia-sdk-python/actions/runs/13545242993

POR-768


TomSPortia (commented):
I think its worth checking with @emmaportia on this one. It leaks quite a bit of internals into a public repo and I know her feelings about the evals being public etc.

emmaportia (dismissed):

emmaportia (approved):

.github/workflows/evals.yml
```diff
@@ -27,14 +27,186 @@ jobs:
...
+          poetry add ../portia-sdk-python
+          poetry install --all-extras
+
+      - name: Create .env.dev file
+        run: |
+          echo "PORTIA_API_KEY=${{ secrets.PORTIA_EVAL_API_KEY }}" > .env.dev
+          echo "PORTIA_API_ENDPOINT=http://0.0.0.0:8080" >> .env.dev
+          echo "POSTGRES_DB=holsten" >> .env.dev
+          echo "POSTGRES_USER=dbuser" >> .env.dev
+          echo "POSTGRES_PASSWORD=dbpassword" >> .env.dev
```
OmarEl-Mohandes, line 80: Probably safer to stores these data in secret keys in github. 

Np: Also, it's nice to add a comment at the beginning saying something like "Consider adding the same changes in the evals repo" so we don't forget to fix stuff in both.

hannah-portia:
Flagging now that this is ready for re-review.
This is migrated to point at the new platform repo and migrated to use staging rather than bringing up a Docker container with a Portia instance.

hannah-portia:
https://github.com/portiaAI/portia-sdk-python/actions/runs/13764930935 <- Here is a successful run I launched manually from the command line. This should take effect once submitted

.github/workflows/evals.yml
```diff
@@ -14,13 +18,133 @@ permissions:
     pull-requests: read
```
OmarEl-Mohandes, line 18: About when it should run. I think we should restrict running evals on specific tag or something. Otherwise, we're just going to spin evals workflow on every commit unnecessarily. 
```
    types:
      - opened
      - synchronize
      - labeled
```

hannah-portia: We had a couple of discussions about this during the standup today.
- we had a bit of pushback about applying labels being extra work
- we also discussed the idea of LLM caching again as reducing the benefit of doing this for cost savings
- the ability to cancel other running actions if a new one is launched was also proposed.

I would like to push forward with this as-is for now. Robbie has a large change that will end up impacting this too and I think its lower priority to restrict the volume of eval runs right now given we have already been doing this in other repos for a few weeks.

Can we revisit in weekly planning?

PR #171, opened 2025-02-26
Fix CLI

robbie-portia:


hannah-portia (dismissed):

portia/cli.py
```diff
@@ -117,9 +118,14 @@ def generate_cli_option_from_pydantic_field(
             if isinstance(info.annotation, type) and issubclass(info.annotation, Enum):
                 field_type = click.Choice(
                     [e.name for e in info.annotation],
-                    case_sensitive=False,
+                    case_sensitive=True,
```
hannah-portia, line 121: Just wanted to double check if this was intended? This means we can't provide lower case inputs?

hannah-portia: Error: Invalid value for '--llm-provider': 'openai' is not one of 'OPENAI', 'ANTHROPIC', 'MISTRALAI'.

robbie-portia: Ah good catch - I thought I did need this when I was going through the debugging, but you're right we don't :) Now updated

hannah-portia (approved):

PR #172, opened 2025-02-27
Fix wait_for_ready to pull workflow once (#160)

OmarEl-Mohandes:
```
cherry-pick b0cc7aec77bb34e681e7a77fa898576ec611dd7b -m 1
```

### DIFF
--- a/portia/runner.py
+++ b/portia/runner.py
@@ -293,7 +293,7 @@ class Runner:
         self.storage.save_workflow(workflow)
         return workflow
 
-    def wait_for_ready(  # noqa: C901
+    def wait_for_ready(
         self,
         workflow: Workflow,
         max_retries: int = 6,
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "portia-sdk-python"
-version = "0.0.2"
+version = "0.2.1"
```
OmarEl-Mohandes, line 3: I'm not sure why do we have 0.0.2 in the first place? .. We need the minor version for fixes, also align with what we [agreed here](https://docs.google.com/document/d/1WLZ0ptgke8gD8GOrPtFQpZ6LwoYg4BD54wmydNZvKQ8/edit?tab=t.0#heading=h.ycrq8npbvi9g)

emmaportia:
> ```
> cherry-pick b0cc7aec77bb34e681e7a77fa898576ec611dd7b -m 1
> ```

what is this voodoo command?

portia/runner.py
```diff
@@ -351,9 +352,6 @@ def wait_for_ready(  # noqa: C901
             # wait a couple of seconds as we're long polling
             time.sleep(backoff_time_seconds)
 
-            # refresh state
-            workflow = self.storage.get_workflow(workflow.id)
```
emmaportia, line 355: if you're deleting this, you should logically be able to delete the if statement on line 358?

emmaportia (commented):
what was the underlying issue? that we no longer need to refresh the workflow because we use the tool ready thing anyway and rely on the SDK to do state updates?

pyproject.toml
OmarEl-Mohandes: got it, will update

portia/runner.py
OmarEl-Mohandes: yea, good point.

OmarEl-Mohandes:
> > ```
> > cherry-pick b0cc7aec77bb34e681e7a77fa898576ec611dd7b -m 1
> > ```
> 
> what is this voodoo command?

Just cherry-picking the commit itself from the main branch. `-m  1`  just to make it work on merge commits, which specifying the main branch itself.

> what was the underlying issue? that we no longer need to refresh the workflow because we use the tool ready thing anyway and rely on the SDK to do state updates?


The underlying issue that caused the bug originally wasn't about the fact that we're reading workflow everytime in the loop, but surfed the bug.  It was because the `current_step_clarifications` had references on clarifications that got "stale", when we override workflow in the loop (because it's initialised once). So when we were resolving `current_step_clarifications` didn't actually resolve the latest workflow clarifications.

emmaportia (approved):

emmaportia (approved):

OmarEl-Mohandes:
I'll cancel this one, as fix is already pushed on main and will get merged today in prod.

PR #180, opened 2025-03-03
Fix clarification imports

robbie-portia:


OmarEl-Mohandes (dismissed):

OmarEl-Mohandes (commented):
You need to bump version if you need this in prod.

robbie-portia:
@OmarEl-Mohandes - thanks, does that look ok?

OmarEl-Mohandes (approved):

portia/__init__.py
```diff
@@ -14,6 +14,8 @@
     MultipleChoiceClarification,
     ValueConfirmationClarification,
 )
+from portia.clarification_handler import ClarificationHandler
+from portia.cli import CLIExecutionHooks
```
Nathanjp91, line 18: I actually don't think this should be included in portia tbh. For people that only want to use this as a library, they shouldn't have to import CLI stuff if they're importing clarifications for example.

robbie-portia: Fair - now removed

Nathanjp91 (approved):

PR #190, opened 2025-03-04
[POR-877] Turnoff summarization for clarifications and enable in search tool

OmarEl-Mohandes:
- Turn off summarisation for clarifications.
- Enable summarisation for the search tool.


emmaportia (approved):
Thanks!

PR #194, opened 2025-03-06
Unroll

TomSPortia:


portia/execution_agents/execution_utils.py
```diff
@@ -135,24 +135,39 @@ def process_output(
...
+    output_values: list[Output] = []
+    for message in messages:
+        if "ToolSoftError" in message.content and tool:
+            raise ToolRetryError(tool.id, str(message.content))
+        if "ToolHardError" in message.content and tool:
+            raise ToolFailedError(tool.id, str(message.content))
+        if clarifications and len(clarifications) > 0:
+            return Output[list[Clarification]](
+                value=clarifications,
+            )
```
OmarEl-Mohandes, line 147: s/return/output_values.append

OmarEl-Mohandes (approved):
Thanks! We definitely need to add evals for this. I agree with you that the prev version was a bit random (of picking only one item). My only concern is that it now gives more control to agents for the tools execution rather than portia instance. 

Overall, i think we should definitely start with this, and see why this might be bad or good. Good catch!

TomSPortia:
> Thanks! We definitely need to add evals for this. I agree with you that the prev version was a bit random (of picking only one item). My only concern is that it now gives more control to agents for the tools execution rather than portia instance.
> 
> Overall, i think we should definitely start with this, and see why this might be bad or good. Good catch!

I actually don't think (though I don't have proof) that it was picking an element at random. I think we we're just parsing the last message returned which makes it seem like it was at random

OmarEl-Mohandes (approved):

PR #197, opened 2025-03-07
Factor out utility method and add test for dynamic Pydantic model generation

sam-portia:
Follow up PR will enhance the model generation slightly, to handle nested types a bit better (we see these in the dummy tools we've added)


robbie-portia (approved):
Looks great 🚀

PR #207, opened 2025-03-10
Use HTTPX client for storage

TomSPortia:


portia/storage.py
```diff
@@ -477,6 +476,15 @@ def __init__(self, config: Config) -> None:
...
         self.api_endpoint = config.must_get("portia_api_endpoint", str)
+        self.client = httpx.Client(
+            base_url=self.api_endpoint,
+            headers={
+                "Authorization": f"Api-Key {self.api_key.get_secret_value()}",
+                "Content-Type": "application/json",
+            },
+            timeout=httpx.Timeout(60),
+            limits=httpx.Limits(max_connections=10),
+        )
```
sam-portia, line 487: Is the change here that makes a difference that we get an httpx client scoped to each concurrent run, or is it just the additional config for the httpx client?

TomSPortia: its actually just the timeout which makes the big difference, but with a client we get connection pooling/reuse which helps.

sam-portia (approved):

PR #210, opened 2025-03-11
Add Notion to excluded by default tools

robbie-portia:
This PR adds Notion to the list of tools that are excluded by default in the tool registry. This ensures that Notion tools are not automatically enabled without explicit setup, as they require OAuth configuration.


robbie-portia:
This was just me playing about - ignore :)

PR #211, opened 2025-03-11
Fix formatting issue when log message has '<>'

fanlinwang:
we will see error like 
```
ValueError: Tag "<repository_name>" does not correspond to any known color directive, make sure you did not misspelled it (or prepend '\' to escape it)
```
if the log message has `<>` in it

Tested with query `star all of the repos belonging to <PortiaAI>`


TomSPortia (approved):
🚀

PR #212, opened 2025-03-12
Fix License Check

TomSPortia:


.github/workflows/licence.yml
```diff
@@ -2,56 +2,57 @@ name: License Check
...
+          python-version: "3.11"
+
+      - name: Install Poetry
+        run: pipx install poetry
+
+      - name: Install dependencies
+        run: |
+          poetry install --all-extras
+          # install license check via pip so it doesn't check itself
+          poetry run pip install licensecheck pipdeptree
```
TomSPortia, line 27: Other than formatting this is the change, using pip to install the checker as `pipdeptree` itself fails license checks!

sam-portia (approved):

PR #234, opened 2025-03-17
Add Introspection agent logic for conditionals

OmarEl-Mohandes:
# Description

- This introducing the Introspection Agent for conditionals. It's an agent that run before executing a step if it has a condition attached to it. 
- The agent is evaluating the outcome of introspection based on PlanRun state and the current plan where it provides outcome as (`CONTINUE`, `SKIP`, `STOP`, `FAIL`).
- Added to the planner a bit of urging to use new `condition` field for conditions if needed.

**Note:** e2e tests are failing due to circular dependency between sdk/backend. Once this is merged, I'll make sure backend is updated with latest and will re-run this step.
 

Ticket Link:  https://linear.app/portialabs/issue/POR-997

## Type of change

(select all that apply)

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

**Example Query:**

**If the weather in London hotter than 10C, sum it with the weather in Cairo and send the result to omar@hello.com**

**Plan:**
```
2025-03-17 17:39:34.431 | DEBUG | portia.portia:plan:210 - Plan: {
    "id": "plan-b627c99d-2e16-4496-9aa4-011cadf384b3",
    "plan_context": {
        "query": "If the weather in London hotter than 10C, sum it with the weather in Cairo and send the result to omar.elmohandes09@gmail.com",
        "tool_ids": [
            "calculator_tool",
            "weather_tool",
            "search_tool",
             ...
        ]
    },
    "steps": [
        {
            "task": "Get the weather in London",
            "inputs": [
                {
                    "name": "$london_city",
                    "value": "London",
                    "description": "City for which to retrieve the weather"
                }
            ],
            "tool_id": "weather_tool",
            "output": "$london_weather",
            "condition": null
        },
        {
            "task": "If London's temperature is above 10C, get the weather in Cairo",
            "inputs": [
                {
                    "name": "$cairo_city",
                    "value": "Cairo",
                    "description": "City for which to retrieve the weather"
                }
            ],
            "tool_id": "weather_tool",
            "output": "$cairo_weather",
            "condition": "$london_weather > 10C"
        },
        {
            "task": "If London's temperature is above 10C, sum the weather in London and Cairo",
            "inputs": [
                {
                    "name": "$london_weather",
                    "value": null,
                    "description": "Temperature or weather value from London"
                },
                {
                    "name": "$cairo_weather",
                    "value": null,
                    "description": "Temperature or weather value from Cairo"
                },
            ],
            "tool_id": "calculator_tool",
            "output": "$weather_sum",
            "condition": "$london_weather > 10C"
        },
        {
            "task": "If London's temperature is above 10C, send an email with the summed weather result",
            "inputs": [
                 ...
                {
                    "name": "$email_body",
                    "value": "The summed weather result is: $weather_sum",
                    "description": "The body of the email containing the weather sum"
                }
            ],
            "tool_id": "portia:google:gmail:send_email",
            "output": "$email_status",
            "condition": "$london_weather > 10C"
        }
    ]
}
```

**PlanRun Output:**

```
{
  "id": "prun-de250d7f-d23c-4456-a6a5-5e1f1504044e",
  "plan_id": "plan-b627c99d-2e16-4496-9aa4-011cadf384b3",
  "current_step_index": 1,
  "state": "COMPLETE",
  "execution_context": {
    "end_user_id": null,
    "additional_data": {},
    "planning_agent_system_context_extension": null,
    "execution_agent_system_context_extension": null
  },
  "outputs": {
    "clarifications": [],
    "step_outputs": {
      "$london_weather": {
        "value": "The current weather in London is overcast clouds with a temperature of 6.69°C.",
        "summary": "The current weather in London is overcast clouds with a temperature of 6.69°C."
      },
      "$cairo_weather": {
        "value": "STOPPED",
        "summary": "The condition for the current step (London's temperature > 10C) is not met as the current temperature is 6.69°C. All remaining steps depend on this condition being true, so execution should stop."
      }
    },
    "final_output": {
      "value": "The current weather in London is overcast clouds with a temperature of 6.69°C.",
      "summary": "London's current temperature is 6.69°C, which is below 10°C. Therefore, the process to sum it with Cairo's weather and send the result to the specified email was not initiated."
    }
  }
}
```
## Changelog

- Added Introspection Agent code.


portia/execution_agents/context.py
```diff
@@ -90,13 +90,10 @@ def generate_input_context(
             "indirectly help you.",
         )
         for output_key in unused_output_keys:
-            # We truncate the output value to 1000 characters to avoid overwhelming the
-            # LLM with too much information.
-            output_val = (str(previous_outputs[output_key].value) or "")[:1000]
             input_context.extend(
                 [
                     f"output_name: {output_key}",
-                    f"output_value: {output_val}",
```
OmarEl-Mohandes, line 99: I reverted back this change of truncating values, because i noticed in various runs not all inputs are correctly included for every step. e.g. `llm_tool` is not providing accurate results of the specific step (extracting something from sheet), while final summary step has the correct result. This issue is exacerbated by conditionals.

TomSPortia: I think this is ok, but lets check in with @robbie-portia given his work on large IO

portia/portia.py
```diff
@@ -493,15 +499,33 @@ def _execute_plan_run(self, plan: Plan, plan_run: PlanRun) -> PlanRun:
 
         dashboard_url = self.config.must_get("portia_dashboard_url", str)
 
+        dashboard_message = (
```
TomSPortia, line 502: Nice!

portia/portia.py
```diff
@@ -568,6 +594,84 @@ def _execute_plan_run(self, plan: Plan, plan_run: PlanRun) -> PlanRun:
...
+            last_executed_step_output (Output | None): The output of the last step executed.
+
+        Returns:
+            tuple[PlanRun, PreStepIntrospectionOutcome]: The updated plan run and the
+                outcome of the introspection.
+
+        """
+        current_step_index = plan_run.current_step_index
+        step = plan.steps[current_step_index]
+        if not step.condition:
```
TomSPortia, line 619: I'm not convinced we want to do this - in my head we always ran the introspection agent because even if there's no condition it may still lead to a FAIL if say the previous step had a bad output?

portia/portia.py
```diff
@@ -493,15 +499,33 @@ def _execute_plan_run(self, plan: Plan, plan_run: PlanRun) -> PlanRun:
...
             plan_run.current_step_index = index
+
+            # Handle the introspection outcome
+            (plan_run, pre_step_outcome) = self._handle_introspection_outcome(
+                introspection_agent=introspection_agent,
+                plan=plan,
+                plan_run=plan_run,
+                last_executed_step_output=last_executed_step_output,
+            )
+            if pre_step_outcome.outcome == PreStepIntrospectionOutcome.SKIP:
```
TomSPortia, line 524: Should we update the last_executed_step_output variable to the output we set in `_handle_introspection_outcome`

portia/portia.py
```diff
@@ -568,6 +594,84 @@ def _execute_plan_run(self, plan: Plan, plan_run: PlanRun) -> PlanRun:
...
+
+        log_message = (
+            f"Pre Introspection Outcome for Step #{current_step_index}: "
+            f"{pre_step_outcome.outcome} for {step.output}. "
+            f"Reason: {pre_step_outcome.reason}",
+        )
+
+        match pre_step_outcome.outcome:
+            case PreStepIntrospectionOutcome.SKIP:
+                logger().debug(*log_message)
```
TomSPortia, line 643: Is there a reason we can't just log on L635? Feels weird to split it like this given we do the same log in each case?

tests/integration/test_e2e.py
```diff
@@ -516,6 +515,19 @@ def on_sleep_called(_: float) -> None:
     assert test_clarification_handler.received_clarification.user_guidance == "please try again"
 
 
+@pytest.mark.flaky(reruns=3)
```
TomSPortia, line 518: OOI what errors do we see that makes this flaky?

portia/introspection_agents/default_introspection_agent.py
```diff
@@ -0,0 +1,91 @@
...
+                        "decide what action should be taken next."
+                        "You should use the current_step_index field to identify the current step"
+                        "in the plan, and the PlanRun state to know what has happened so far."
+                        "The actions that can be taken next in priority order are:"
+                        " - STOP -> stop execution and return the result so far."
+                        " - SKIP -> skip the next step execution."
+                        " - FAIL -> stop execution entirely."
+                        " - CONTINUE -> Continue execution of the next step."
+                        "You should choose an outcome based on the following logic in order:\n"
+                        " - If all remaining steps can not be executed because of the"
```
TomSPortia, line 54: I'm not sure this is quite correct? We should STOP because its done right?

```suggestion
                        " - If the overarching goal of the plan has already been met return SKIP"
```

portia/introspection_agents/default_introspection_agent.py
```diff
@@ -0,0 +1,91 @@
...
+                        "   current conditions are not met return STOP.\n"
+                        " - If the current step has a condition that is false you return SKIP.\n"
+                        " - If you cannot evaluate the condition because some data had been skipped"
```
TomSPortia, line 57: Likewise I think if you can't evaluate a condition it should be FAIL not SKIP

TomSPortia (commented):
Looking great, some small things but I think we should probably workshop the prompt of the agent a little to make sure we're super clear when we're getting SKIP/FAIL/CONTINUE etc

portia/introspection_agents/default_introspection_agent.py
OmarEl-Mohandes: I don't think so, we want to support chain conditionals cases. Consider a case where:

step1 = if (x > 10) fetch github repo. ==> SKIP 
step2 = If (repo is found) do y.  ==> SKIP not FAIL.
step3 = ...

TomSPortia: ah but to me thats not I can't evaluate a condition, thats the condition being evaluated. understand the LLM may not view it that way though

tests/integration/test_e2e.py
OmarEl-Mohandes: I'm trying play with it a bit. 

Two cases I found: 
  - Sometimes it's the same output names of the 2 branches, which fails the plan itself. (This happens on different models sometimes  e.g. 4o-mini).
  - Sometimes it's the accuracy of the condition evaluation is wrong, and does just return just temprature. I'm hoping to write a bit hard evals on this.

portia/portia.py
OmarEl-Mohandes: I've split it for becuase I'm erroring in FAIL case.

portia/introspection_agents/default_introspection_agent.py
OmarEl-Mohandes: Yea, LLM was very aggresive on this specific case. If one condition got skipped, that leads to different step it'll just say I can't evaluate fail.

portia/introspection_agents/default_introspection_agent.py
OmarEl-Mohandes: Yea, happy to give it a try. I'm a bit skiptical on doing it overall all goal, as I'm afraid of hallucinating and just stops prematurely.

portia/portia.py
OmarEl-Mohandes: Thought about this, but what I wanted is to have this on last_executed (as if tool executed) output, rather than shortcutting using agent. Mainly, I need it for constructing the final_output value. The final output summary is going to handle skimming over all outputs anyways.

portia/portia.py
OmarEl-Mohandes: good point, let me try to re-iterate the prompt a bit and test. Happy to remove this.

portia/execution_agents/context.py
OmarEl-Mohandes: Discussed offline, I'll revert it back with more relaxed limit and Robbie will remove it once we introduced big I/O handling.

portia/portia.py
OmarEl-Mohandes: I found one case where this might prematurally fail the plan run, which is input clarifications. 
So interospection is returning FAIL (becuase it's true, it should fail), but we handle that afterwards through clarifications.

portia/plan.py
```diff
@@ -295,16 +301,15 @@ def __str__(self) -> str:
...
-        Checks that the plan has at least one step, that all outputs are unique, and that all
-        steps use valid outputs as inputs.
+        Checks that all outputs + conditions are unique.
 
         Returns:
             Plan: The validated plan.
 
         """
-        outputs = [step.output for step in self.steps]
+        outputs = [step.output + (step.condition or "") for step in self.steps]
```
TomSPortia, line 310: Won't this break if there are multiple steps with no condition? Also reflecting on it isn't it ok for conditions to be duplicated. Like a query like: `send an email to tom if the temp > 10 and also send one to emma if the temp > 10` is valid?

portia/introspection_agents/default_introspection_agent.py
```diff
@@ -0,0 +1,90 @@
...
+                        "You should use the current_step_index field to identify the current step"
+                        "in the plan, and the PlanRun state to know what has happened so far."
+                        "The actions that can be taken next are:"
+                        " - STOP -> stop execution and return the result so far."
+                        " - SKIP -> skip the next step execution."
+                        " - FAIL -> stop execution entirely."
+                        " - CONTINUE -> Continue execution of the next step."
+                        "You should choose an outcome based on the following logic in order:\n"
+                        " - If the overarching goal of the plan has already been met return STOP.\n"
+                        " - If the current step has a condition that is false you return SKIP.\n"
```
TomSPortia, line 55: I still think we should add the FAIL if not possible condition, but I don't want to block us moving forward so if you strongly believe we shouldn't lets add a ticket to explore ways of extending this to handle this case:

```
- If the next step cannot be completed because of the outputs of previous steps return FAIL
```

TomSPortia (approved):
Looking good, one small question but then I think we can get it in

portia/plan.py
OmarEl-Mohandes: > Won't this break if there are multiple steps with no condition?

I don't think so. outputs with no conditions should be treated as usual (just outputs). It'll only break if the same output_name is used multiple times with no conditions, which i think in this case it should error it out.

> Also reflecting on it isn't it ok for conditions to be duplicated...

I tried this query for 10 times in a row (and even the same email used on both statements), the output name are always different in both steps (with the same condition), which is expected because planner has that statement of making sure outputs are unique.

portia/introspection_agents/default_introspection_agent.py
TomSPortia: imagine a query: 

```
find the portialabs blog post about dogs, then favourite it
```

where the first tool returns [] for the search. At the moment, unless the planner is smart enough to add a condition to the favourite step (like $blog_post is not null), then we'll try and run the favourite step. 

Whereas we should fail the workflow if there no blog post found

portia/plan.py
TomSPortia: oh sorry misread the code, I though it was [step.output, step.condition] not [step.output + step.condition]. 

Still what about if step.output is undefined? Like a plan like:

```
{steps: [{condition: None, output: None},{condition: None, output: None}]}
```

i think this is legit right? I can have a plan that just calls two calls but doesn't need the output?

portia/introspection_agents/default_introspection_agent.py
OmarEl-Mohandes: gotcha, I agree with you I'll create a ticket for it. The main challenge we're facing here is that introspection agent might think it can't meet this goal, where a clarification later might resolve it. 

I've created a ticket for it: https://linear.app/portialabs/issue/POR-1045

portia/plan.py
OmarEl-Mohandes: I've never seen a step without an output, it's required field. Even if it's just a call (like send an email). It's has an output as $email_sent.

portia/introspection_agents/default_introspection_agent.py
TomSPortia: yeah fair. it may be that we want to clarify rather than fail in that case. Lets ticket

portia/plan.py
TomSPortia: ok yeah, it seems like the planner always adds one, even if its not needed. Happy to go with this

PR #235, opened 2025-03-18
Add tracing to claude evals

robbie-portia:
# Description

Add tracing to claude evals now that they've added a wrap_anthropic() method (we also needed to update langsmith to access it)

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update


portia/llm_wrapper.py
```diff
@@ -223,8 +223,10 @@ def to_instructor(
                 )
             case LLMProvider.ANTHROPIC:
                 client = instructor.from_anthropic(
-                    client=Anthropic(
-                        api_key=self.api_key.get_secret_value(),
+                    client=wrappers.wrap_anthropic(
+                        Anthropic(
+                            api_key=self.api_key.get_secret_value(),
+                        ),
```
sam-portia, line 229: Yesterday I saw the wrapper for OpenAI when I was looking at Azure OpenAI and I did wonder if this was why we didnt see traces!

sam-portia (approved):
👌

portia/llm_wrapper.py
robbie-portia: Yeah, when this was first added there wasn't an equivalent wrapper for anthropic but that was ok because evals weren't running on anthropic

robbie-portia:
Note: I'll merge once https://github.com/portiaAI/platform/pull/109 is in - I think that is needed for the evals to pass here

PR #242, opened 2025-03-19
Plan with cloud tools on error

sam-portia:
# Description

Based on user feedback

If you haven't got an API key set then the planner can fail when building more than a trivial plan because the tools available are limited.

This PR introduces a log message that only gets triggered if:
- the user has no API key
- the Plan generation fails
- the replan with Portia cloud tools does not fail

[Ticket Link](https://linear.app/portialabs/issue/POR-829)

## Type of change

(select all that apply)

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [x] Documentation update

## Screenshots

```python
from portia import Portia, Config

portia = Portia(config=Config(portia_api_endpoint="https://api.porita.dev"))
portia.run("Send an email to example@example.com about farm yard animals")
```

Results in 
image

## Changelog

### DIFF
--- a/portia/portia.py
+++ b/portia/portia.py
@@ -709,22 +709,18 @@ class Portia:
         example_plans: list[Plan] | None = None,
     ) -> None:
         """Generate a plan using Portia cloud tools for users who's plans fail without them."""
-        try: # This steps is optional and if it fails we don't want to exit with an error
-            cloud_registry = (
-                self.tool_registry
-                + PortiaToolRegistry.with_default_tool_filter(self.config)
-            )
-            tools = cloud_registry.match_tools(query)
-            planning_agent = self._get_planning_agent()
-            replan_outcome = planning_agent.generate_steps_or_error(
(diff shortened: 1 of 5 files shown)


portia/portia.py
```diff
@@ -690,3 +701,34 @@ def _get_agent_for_step(
...
+        original_error: str,
+        query: str,
+        example_plans: list[Plan] | None = None,
+    ) -> None:
+        """Generate a plan using Portia cloud tools for users who's plans fail without them."""
+        cloud_registry = (
+            self.tool_registry
+            + PortiaToolRegistry.with_default_tool_filter(self.config)
+        )
+        tools = cloud_registry.match_tools(query)
```
OmarEl-Mohandes, line 716: I might have missed this, but does `get_tools` from cloud works without api key?

OmarEl-Mohandes (approved):

PR #246, opened 2025-03-21
Move mistral to optional / extras dependency

sam-portia:
# Description

OpenAI and Anthropic are left as "default" for now - thoughts?

Added a guard for imports to direct people to install if they try to use Mistral without the extra dep

Usage: 
`poetry install portia-sdk-python[mistral]`

For local development:
`poetry install --all-extras`

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [x] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/llm_wrapper.py
+++ b/portia/llm_wrapper.py
@@ -26,10 +26,8 @@ from typing import TYPE_CHECKING, TypeVar
 import instructor
 from anthropic import Anthropic
 from langchain_anthropic import ChatAnthropic
-from langchain_mistralai import ChatMistralAI
 from langchain_openai import ChatOpenAI
 from langsmith import wrappers
-from mistralai import Mistral
 from openai import OpenAI
 from pydantic import BaseModel, SecretStr
 
@@ -184,6 +182,8 @@ class LLMWrapper(BaseLLMWrapper):
                     api_key=self.api_key,
                 )
             case LLMProvider.MISTRALAI:
+                from langchain_mistralai import ChatMistralAI
+
                 return ChatMistralAI(
                     model_name=self.model_name.value,
                     api_key=self.api_key,
@@ -237,6 +237,8 @@ class LLMWrapper(BaseLLMWrapper):
                     max_tokens=2048,
                 )
             case LLMProvider.MISTRALAI:
+                from mistralai import Mistral
+
                 client = instructor.from_mistral(
                     client=Mistral(
                         api_key=self.api_key.get_secret_value(),
--- a/portia/llm_wrapper.py
+++ b/portia/llm_wrapper.py
@@ -31,8 +31,7 @@ from langsmith import wrappers
 from openai import OpenAI
 from pydantic import BaseModel, SecretStr
 
-from portia.common import is_library_installed
-from portia.config import Config, LLMModel, LLMProvider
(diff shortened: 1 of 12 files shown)


portia/llm_wrapper.py
```diff
@@ -237,6 +243,13 @@ def to_instructor(
                     max_tokens=2048,
                 )
             case LLMProvider.MISTRALAI:
+                if not is_library_installed("mistralai"):
```
Nathanjp91, line 246: if we're doing this for multiple groups, rather than repeating this pattern a bunch of places, we might just want to do something like below, moving the raise with custom message to the function check

```python
def extra_group_enabled(extra_group: str) -> None:
    match extra_group:
        case "mistralai"
            if not importlib.util.find_spec("mistralai") and not importlib.util.find_spec("langchain-mistralai"):
                 raise ImportError("MistralAI group install message")
        ...
```

Nathanjp91: I can't imagine a case where we'd want to check but not then try importing, so may as well keep the other sections of our code a bit tighter. But if we do want to just check, the above could return a bool and have a should_raise=True input to the function that triggers the importerror

Nathanjp91: means we tie the extra groups to this function, but then we don't have to update it everywhere if we change an extra group

Nathanjp91: this might be a list[str] here as well if we have functonality that requires multiple extra groups

Nathanjp91 (changes requested):
Like it, but do think there might be a more long term maintanable and extensible approach. Let you consider though. Happy to hear if you disagree.

portia/llm_wrapper.py
sam-portia: What about:

```python
EXTRAS_GROUPS_DEPENDENCIES = {
    "mistral": ["mistralai", "langchain-mistralai"],
    ...
}

def validate_extras(extra_group: str) -> None:
    if not all(importlib.util.find_spec(p) for p in EXTRAS_GROUPS_DEPENDENCIES[extra_group]):
        raise ImportError(f"Please install portia-sdk-python[{extra_group}] to use these models")
```

Or too much?

sam-portia: Slightly irks me that its a duplicate of the pyproject.toml 😐

Nathanjp91: I don't think there's anyway around that if we wanted this approach, just have to maintain the 2 lists, one in code and one in toml file. The other approach that could be taken is on the very top level package init file, just do like

```python
ENABLED_EXTRAS = {}
try:
    import mistralai
    import langchain_mistralai
except:
    ENABLED_EXTRAS.append('mistral')
```
Then validate_extras just becomes a check against that list or raise with custom message if not.

Either way though, we're stuck with maintaining 2 sources here. Maybe write a custom github action to keep them in line if feel strongly.

Nathanjp91: Like the approach with validate_extras you've got above though, very easy to maintain

Nathanjp91 (approved):

PR #249, opened 2025-03-22
Local browser tool

emmaportia:
# Description

Add the first local version of the browser tool within experimental / future directory.
- This also sets up tool specific dependencies to be installed using the poetry [feature] syntax, e.g poetry add portia-sdk-python[tools-browser-local]

Ticket Link: POR-1018

## Type of change

(select all that apply)

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update


portia/_experimental/_future/_unstable/browser_tool.py
```diff
@@ -0,0 +1,184 @@
...
+        description="The URL to navigate to for login if the user is not authenticated.",
+    )
+    user_login_guidance: str | None = Field(
+        default=None,
+        description="Guidance for the user to login if they are not authenticated.",
+    )
+
+
+class BrowserTool(Tool[str]):
+    """General purpose browser tool. Customizable to user requirements."""
```
emmaportia, line 73: I'll expand on this a lot before finalising, but would rather spend the time on that when both the browser local and browserbase versions are finished.

Nathanjp91:
I'll review fully later, but I don't think we need 3 tiers of folder here. I meant more just picking a name and then sticking with that. I like _unstable personally, and then just keeping everything in that folder. But not fussed which we pick. If we want to get really optimal then we mirror the end structure in feature subfolders. Like portia._unstable.browser.{folder structure reflection}, that way you can just drop the files from feature directly into the main repo when it's done. But that is an unnecessary nicety. I'd just keep it top level if they're small features.

PR #251, opened 2025-03-23
[POR-1003] Custom agents

emmaportia:
# Description

In order to have our own planning agents that we are testing and refining, we need to be able to use these in the SDK without forking the code. We should probably also extend the concept to execution agents.

Replaces https://github.com/portiaAI/portia-sdk-python/pull/223

Re the discussion on the other PR about having this as a separate method on the config class -- I think that creates another way to do config like actions so I'd rather not do that unless the reason for it is strong. At the moment we don't have a requirement for serialisable config so I'm not sure if it would be the right tradeoff but happy to discuss further.
@TomSPortia @Nathanjp91 -- plz review directionally (hence the draft PR)

## Type of change

(select all that apply)

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

### Added
- 0.1.8 Allow for custom execution and planning agents without forking of code

### DIFF
--- a/portia/config.py
+++ b/portia/config.py
@@ -170,6 +170,7 @@ class ExecutionAgentType(Enum):
 
     ONE_SHOT = "ONE_SHOT"
     DEFAULT = "DEFAULT"
+    CUSTOM = "CUSTOM"
 
 
 class PlanningAgentType(Enum):
@@ -181,6 +182,7 @@ class PlanningAgentType(Enum):
     """
 
     DEFAULT = "DEFAULT"
+    CUSTOM = "CUSTOM"
 
 
 class LogLevel(Enum):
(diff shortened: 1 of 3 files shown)


portia/config.py
```diff
@@ -25,6 +25,15 @@
 
 from portia.errors import ConfigNotFoundError, InvalidConfigError
 
+if TYPE_CHECKING:
+    from portia.execution_agents.base_execution_agent import BaseExecutionAgent
+    from portia.planning_agents.base_planning_agent import BasePlanningAgent
+else:
+    # Placeholder types for runtime
+    BaseExecutionAgent = Any
```
Nathanjp91, line 33: is it a circular dependency otherwise? I've never seen this before. If it's only used for type checking it should be fine to do without.

portia/config.py
```diff
@@ -425,7 +447,23 @@ def parse_execution_agent_type(cls, value: str | ExecutionAgentType) -> Executio
...
-        return parse_str_to_enum(value, PlanningAgentType)
+        planning_agent_type = parse_str_to_enum(value, PlanningAgentType)
+        if planning_agent_type == PlanningAgentType.CUSTOM and not cls.custom_planning_agent:
+            raise InvalidConfigError(
+                "planning_agent_type",
+                "Custom planning agent not set",
+            )
+        return planning_agent_type
+
+    custom_planning_agent: type[BasePlanningAgent] | None = Field(
```
Nathanjp91, line 458: would they be using multiple planning agents? Why not just overwrite config.planning_agent?

portia/config.py
```diff
@@ -299,7 +312,10 @@ class Config(BaseModel):
 
     """
 
-    model_config = ConfigDict(extra="ignore")
+    model_config = ConfigDict(
+        arbitrary_types_allowed=True,
+        populate_by_name=True,
```
Nathanjp91, line 317: is the intention to give the agents alias's?

emmaportia (commented):
Thanks!

portia/config.py
Nathanjp91: I don't feel that great about this. I'm happy to merge this for now if we put a ticket in the backlog to decouple config and agents. Looking at the agents and config atm, we don't really need to be passing in the config, only select peices of information to get the llm needed. Which we could either do ahead of time and just pass in, or pass in the specific information required to build it inside the agent. Then agents and config would not be a circular dependency.

Nathanjp91 (approved):

PR #257, opened 2025-03-25
Fix docs

emmaportia:
# Description

Exclude the _unstable directory from docs generation

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,7 +121,7 @@ type = "python"
 
 [[tool.pydoc-markdown.processors]]
 type = "filter"
-expression = "not 'test' in name and not 'cli' in name and not 'prefixed_uuid' in name and not 'common' in name and not 'templates' in name and default()"
+expression = "not 'test' in name and not 'cli' in name and not 'prefixed_uuid' in name and not 'common' in name and not 'templates' in name and not '_unstable' in name and default()"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "portia-sdk-python"
-version = "0.1.9-alpha"
+version = "0.1.10-alpha"
```
sam-portia, line 3: Should be 0.1.9-alpha2 ? Not sure exactly on the syntax

emmaportia: I don't think we actually need to bump the version to reflect this change, I double checked the github action.

But no, I think we agreed to bump the minor version rather than having a trailing value
https://docs.google.com/document/d/1WLZ0ptgke8gD8GOrPtFQpZ6LwoYg4BD54wmydNZvKQ8/edit?tab=t.0#heading=h.z71o1ec9h3y0

sam-portia: Do you want to revert the revert to get this change back in then?

sam-portia (approved):
No problem for getting merged but think it needs the 0.1.10-alpha version back before the pypi push

PR #266, opened 2025-03-31
WIP Remove LLMWrapper Part 1: Gut the class

sam-portia:
# Description

Goal of this PR is to move towards the removal of LLMWrapper. Ability to configure new Model classes more directly will be in a follow up.

In this PR:
* LLMWrapper is a thin wrapper around a Model
* Required changes to `DefaultExecutionAgent` are made (because we had a separate implementation of structured output)
* Tests are updated to use a new Mock object which utilises unittest.mock objects rather than our own custom class

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [x] Refactor
- [ ] Documentation update

## Screenshots

## Changelog


portia/execution_agents/execution_utils.py
```diff
@@ -177,82 +171,3 @@ def process_output(  # noqa: C901
         summaries.append(output.summary or output.serialize_value(output.value))
 
     return Output(value=values, summary=", ".join(summaries))
-
-
-def map_message_types_for_instructor(
```
sam-portia, line 182: This logic that was introduced temporarily is not part of the Model type

tests/unit/execution_agents/test_default_execution_agent.py
```diff
@@ -122,32 +74,31 @@ def test_parser_model(monkeypatch: pytest.MonkeyPatch) -> None:
...
     assert messages
     assert "You are a highly capable assistant" in messages[0].content  # type: ignore  # noqa: PGH003
     assert "CONTEXT_STRING" in messages[1].content  # type: ignore  # noqa: PGH003
     assert "DESCRIPTION_STRING" in messages[1].content  # type: ignore  # noqa: PGH003
     assert "TOOL_NAME" in messages[1].content  # type: ignore  # noqa: PGH003
     assert "TOOL_DESCRIPTION" in messages[1].content  # type: ignore  # noqa: PGH003
     assert "INPUT_DESCRIPTION" in messages[1].content  # type: ignore  # noqa: PGH003
-    assert mock_invoker.output_format == ToolInputs
+    assert mock_model.with_structured_output.called
+    assert mock_model.with_structured_output.call_args[0][0] == ToolInputs
```
sam-portia, line 93: The refactors in this test are duplicated many times below

portia/config.py
```diff
@@ -438,6 +449,49 @@ def model(self, usage: str) -> LLMModel:
...
+    def resolve_model(self, usage: str) -> GenerativeModel:
+        """Resolve a model from the config."""
+        if usage in self.custom_models:
+            return self.custom_models[usage]
+        model = self.model(usage)
+        return self._construct_model(model)
+
+    def _construct_model(self, llm_model: LLMModel) -> GenerativeModel:
+        """Construct a Model instance from an LLMModel."""
+        match llm_model.provider():
```
Nathanjp91, line 461: I assume that this is just an interim step on the way to removing this llm_model class? Like the end goal is to have this _construct_model call liteLLM with some config options?

portia/config.py
```diff
@@ -537,21 +591,10 @@ def validate_llm_api_key(provider: LLMProvider) -> None:
 
         validate_llm_api_key(self.llm_provider)
         for model in self.models.values():
-            validate_llm_api_key(model.provider())
-
+            if isinstance(model, LLMModel):
+                validate_llm_api_key(model.provider())
```
Nathanjp91, line 595: Is kind of a stretch goal and bit of a small one, but as part of the llm angosticism work I'd love to see us just store llm api keys either as single field (rather than specific ones like .openai_api_key). Something like 

```python
class Config:
    api_keys: dict[str, str]
```

Where the class loader just recognizes the models you've got by name, say if you're doing litellm style 'openai/o3-mini', it just detects that it's an openai model and loads the dictionary with {"openai": os.get.env("OPENAI_API_KEY"). 

I think this makes key management easier and more scaleable towards supporting different providers as well.

portia/model.py
```diff
@@ -443,6 +441,7 @@ def get_structured_response_instructor(
 
 
 if validate_extras_dependencies("mistral", raise_error=False):
+    from langchain_mistralai import ChatMistralAI
     from mistralai import Mistral
 
     class MistralAIGenerativeModel(LangChainGenerativeModel):
```
Nathanjp91, line 447: This technique is interesting. Do these show up in importable items? Like from another module will vscode give me errors in imports if I try to do `from portia.model import MistralAIGenerativeModel`?

Nathanjp91 (approved):
LGTM: Nothing blocking as I can see.

portia/config.py
sam-portia: Yes it's interim. I think we should maintain LLMModel / LLMProvider as ways to configure for at least 1 more release before removing?

portia/config.py
sam-portia: Yep I've got that on the radar to consider today

portia/model.py
sam-portia: They do in development because we'll always have all extras installed I think? I would assume if the extra isnt installed then it wouldn't show up, but maybe VSCode does something clever. Will try it out

sam-portia: VSCode isnt giving me errors - it resolves the symbol even if the extras-validation fails 

When I actually import it gives an ImportError, as expected.

I'll see how this feels after thinking a bit more about the config, I think atm its OK

Nathanjp91: Wasn't a blocker, more just a casual note of behaviour I didn't know the import modules were smarter than just top level classes. I wonder how deep it can go.

PR #270, opened 2025-04-01
Sort large outputs in agent memory

robbie-portia:
# Description

* Set up an agent-memory feature flag which defaults to false (for now). Without this, there shouldn't be any changes to behaviour.
* In the portia instance, when updating outputs we now save them to agent memory if they are large and memory is turned on.
* Add agent memory to the storage classes (local, disk and cloud) to allow putting and getting outputs from agent memory
* Separated the outputs into their own file and allowed for values that are stored in agent memory (either locally, remote or on disk). It now has two ways of accessing the value: value_for_prompt() which gives a version that fits into a prompt, and full_value() which retrieves the full value.
* Make it so that we always summarise large outputs, so that we have a value to work with locally. In order to do this, the summariser summarises just based on the first part of the output.

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog


portia/cloud.py
```diff
@@ -18,17 +18,25 @@ def get_client(cls, config: Config) -> httpx.Client:
         return cls._client
 
     @classmethod
-    def new_client(cls, config: Config, *, allow_unauthenticated: bool = False) -> httpx.Client:
+    def new_client(
+        cls,
+        config: Config,
+        *,
+        allow_unauthenticated: bool = False,
+        json_headers: bool = True,
```
robbie-portia, line 26: This is needed because we're sending a multi-part form for the output upload (rather than json), so we need to not set `application/json`.

Nathanjp91 (commented):
Some comments for consideration

Nathanjp91:
> One question I do have - what's the best way to roll this out? The platform depends on the 'Output' type when running the evals (error message below). Should I re-export from portia.execution_agents.base_execution_agent until this makes it out to production, then move the evals over to the new import and undo the re-export?
> 
> `ImportError: cannot import name 'Output' from 'portia.execution_agents.base_execution_agent' `
> 
> CC @Nathanjp91 / @sam-portia

It'll need to be updated on the platform, you can pin the evals + backend with specific commit hashes, including open PR's, multistage release for staging basically, change platform to be pinned to the commit, test against that and validate, then merge the sdk, and swap the platform back to latest main, or an alpha release.

portia/storage.py
```diff
@@ -690,3 +846,76 @@ def save_tool_call(self, tool_call: ToolCallRecord) -> None:
         else:
             self.check_response(response)
             LogAdditionalStorage.save_tool_call(self, tool_call)
+
+    def save_plan_run_output(
```
sam-portia, line 850: Should we save the outputs to disk and only get them from the cloud if they have disappeared?

We can / possibly should store a checksum of their content in our backend so we can verify its the same data locally

sam-portia: We might actually want to save the contents in the DB with the checksum being the primary key, so that we don't redundantly store information...

tests/unit/execution_agents/test_output.py
```diff
@@ -0,0 +1,89 @@
...
+        (1, "1"),
+        (1.23, "1.23"),
+        (False, "false"),
+        (LLMModel.GPT_4_O, str(LLMModel.GPT_4_O.value)),
+        (MyModel(id="123"), MyModel(id="123").model_dump_json()),
+        (b"Hello World!", "Hello World!"),
+        (now, now.isoformat()),
+        (not_a_model, str(not_a_model)),
+        ([clarification], json.dumps([clarification.model_dump(mode="json")])),
+    ]
```
sam-portia, line 55: [This should probably be a `@pytest.mark.parametrize`](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html)

tests/unit/execution_agents/test_output.py
```diff
@@ -0,0 +1,89 @@
...
+def test_agent_memory_output() -> None:
+    """Test value is stored in agent memory."""
+    storage_details = AgentMemoryStorageDetails(
+        name="test_value",
+        plan_run_id=PlanRunUUID(),
+    )
+    output = Output(value=storage_details, summary="test summary")
+    assert output.value_for_prompt() == "test summary"
+
+    mock_agent_memory = MagicMock()
```
sam-portia, line 81: You could probably set the spec here `MagicMock(spec=AgentMemory)` so we validate attributes and methods correctly

tests/unit/execution_agents/utils/test_step_summarizer.py
```diff
@@ -78,6 +87,46 @@ def test_summarizer_model_no_messages() -> None:
     assert result["messages"] == [None]
 
 
+def test_summarizer_model_large_output() -> None:
+    """Test the summarizer model with large output."""
+    summary = AIMessage(content="Short summary")
+    mock_invoker = get_mock_base_chat_model(response=summary)
```
sam-portia, line 93: warning this will change with my PR!

tests/unit/test_portia.py
```diff
@@ -545,6 +557,88 @@ def test_portia_sets_final_output_with_summary(portia: Portia) -> None:
...
+
+    mock_summarizer_agent = mock.MagicMock()
+    mock_summarizer_agent.create_summary.side_effect = [expected_summary]
+
+    with (
+        mock.patch.object(
+            LLMWrapper,
+            "to_instructor",
+            new=MagicMock(return_value=mock_plan),
+        ),
```
sam-portia, line 596: also will change!

robbie-portia: Thanks for the heads up - now updated!

tests/unit/execution_agents/utils/test_step_summarizer.py
robbie-portia: Thanks for the heads up - now updated!

tests/unit/execution_agents/test_output.py
robbie-portia: Ah cool, I didn't realise this was a thing - nice!

tests/unit/execution_agents/test_output.py
robbie-portia: Agreed - I was just copying this across but should've tidied it up while doing that (or more realistically, asked cursor to!). Now done.

portia/portia.py
```diff
@@ -880,3 +882,9 @@ def _log_replan_with_portia_cloud_tools(
 
     def _get_introspection_agent(self) -> BaseIntrospectionAgent:
         return DefaultIntrospectionAgent(self.config)
+
```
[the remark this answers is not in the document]
robbie-portia: Yeah, agree this isn't ideal, but was also a bit nervous about trying to make bigger changes to the way we store state as part of this PR. I've had a bit of a go at adding something like this that moves us in the right direction. Let me know what you think.

portia/storage.py
robbie-portia: We could do, though I have to admit that I think something like this might be a premature optimisation. I think it's worth getting this as-in for now, then we could add local caching pretty easily if we find we need it.

portia/execution_agents/output.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class AgentMemoryStorageDetails(BaseModel):
+    """Details about the storage of an output in agent memory."""
+
+    name: str
+    plan_run_id: PlanRunUUID
+
+
+class Output(BaseModel, Generic[SERIALIZABLE_TYPE_VAR]):
```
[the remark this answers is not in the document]
robbie-portia: Yeah, I chickened out of doing this first time around because I thought it might be a v big change lol. This nudged me into having another go and I think I now have something that looks ok with them as two separate classes. What do you think?

portia/cloud.py
robbie-portia: Yeah, I completely agree. Mind if I raise a ticket for it and put it in a future cycle?

portia/storage.py
sam-portia: Im not so sure, but it probably depends on the implementation of the execution agent. Something iterative that is reading / writing large files over the network at every step feels potentially problematic? As soon as we have one workflow that has one large output in it, my understanding is that every step that runs after that will (or could) re-download that large output, is that right? 

Part of the reason I suggested it is that it feels like a low effort / high reward addition (high reward in the sense that I would assume this cache would have a 100% hit rate). But happy to be overridden on either the effort/reward hypothesis

sam-portia (approved):
Looks good - no real blockers although I think the simple filesystem cache is worth pursuing unless prohibitive in some way!

portia/execution_agents/output.py
```diff
@@ -0,0 +1,129 @@
...
+    def serialize_value(self) -> str:
+        """Serialize the value to a string."""
+        return self.summary
+
+    def full_value(self, agent_memory: AgentMemory) -> Serializable | None:
+        """Get the full value, fetching from remote storage or file if necessary."""
+        return agent_memory.get_plan_run_output(self.output_name, self.plan_run_id)
+
+
+Output = Union[LocalOutput, AgentMemoryOutput]
```
sam-portia, line 129: It miiight be worth having a common base class `Output` here so that we're forced to keep the common methods in sync (as it is `AgentMemoryOutput.full_value` could drift from `LocalOutput.full_value` for example)

robbie-portia: So I tried this initially, and it was causing issues with pydantic serialisation. Essentially, if plan run stored an Output and was serialised and then de-serialised, it wouldn't know how to de-serialise the abstract Output into a concrete type. I tried a few different things with discriminators but it got quite ugly and I couldn't get it to work nicely. Any ideas how to get around this?

sam-portia: Oh I see - yeah descriminators is the right approach for dealing with unions on deserialisation

You can make the base-class _not_ a pydantic BaseModel - as in

```python
class Output:
    @abstractmethod
    def full_value(self, ...):
        ...

class LocalOutput(Output, pydantic.BaseModel):
    ... # pydantic fields etc
    def full_value(self, ...):
        ...
```

That might get around your issues

Nathanjp91: The pydantic fix there is that you set the ordering of deserialisation on the PlanRun class. OutputClass1 | OutputClass2, it'll first try to deserialise the value to type 1, then if it fails, to type 2. You could also set it as Output, and then write a custom deserialiser function for that field on PlanRun that returned the right type.

portia/storage.py
```diff
@@ -214,35 +221,109 @@ def save_tool_call(self, tool_call: ToolCallRecord) -> None:
...
+        """
+        converted_output = output.model_copy()
+        converted_output.value = AgentMemoryStorageDetails(
+            name=output_name,
+            plan_run_id=plan_run_id,
+        )
+        return converted_output
+
+
+class InMemoryStorage(PlanStorage, RunStorage, LogAdditionalStorage, AgentMemory):
```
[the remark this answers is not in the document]
Nathanjp91: I think it's probably fine for now. We can revisit it later if we come up against issues.

portia/cloud.py
robbie-portia: Created POR-1206

portia/execution_agents/output.py
```diff
@@ -19,7 +20,33 @@
...
+
+        This value may be long and so is not suitable for use in LLM prompts.
+        """
+
+    @abstractmethod
+    def get_summary(self) -> str | None:
+        """Return the summary of the output."""
+
+
+class LocalOutput(BaseOutput, Generic[SERIALIZABLE_TYPE_VAR]):
```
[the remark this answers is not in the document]
robbie-portia: Unfortunately you can't - to be honest, I'm not 100% sure why, but the linter complains

portia/storage.py
```diff
@@ -626,15 +615,77 @@ def get_plan_run_output(self, output_name: str, plan_run_id: PlanRunUUID) -> Out
 class PortiaCloudStorage(Storage, AgentMemory):
     """Save plans, runs and tool calls to portia cloud."""
 
-    def __init__(self, config: Config) -> None:
+    MAX_CACHE_SIZE = 20
```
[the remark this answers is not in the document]
robbie-portia: Done

PR #281, opened 2025-04-03
Refactor ToolRegistry to simplify code. Add model field to LLMTool, Image tool and browser tool

sam-portia:
# Description

Step 1 of simplifying config: Allow user to provide `GenerativeModel` to `LLMTool`, `ImageUnderstandingTool` and `BrowserTool`

Simplified ToolRegistry based on observation that the only real difference between different implementations is the instantiation, the rest is the same.

Ticket Link: [POR-1143](https://linear.app/portialabs/issue/POR-1143/simplify-config)

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [x] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/tool_registry.py
+++ b/portia/tool_registry.py
@@ -18,7 +18,6 @@ from __future__ import annotations
 import asyncio
 import os
 import re
-from abc import ABC, abstractmethod
 from enum import StrEnum
 from typing import TYPE_CHECKING, Any, Callable, Literal, Union
 
@@ -52,16 +51,19 @@ if TYPE_CHECKING:
     from portia.config import Config
 
 
-class ToolRegistry(ABC):
-    """ToolRegistry is the base interface for managing tools.
+class ToolRegistry:
+    """ToolRegistry is the base class for managing tools.
 
-    This class defines the essential methods for interacting with tool registries, including
-    registering, retrieving, and listing tools. Specific tool registries should implement these
-    methods.
+    This class implements the essential methods for interacting with tool registries, including
+    registering, retrieving, and listing tools. Specific tool registries can override these methods
+    and provide additional functionality.
 
     Methods:
-        register_tool(tool: Tool) -> None:
-            Registers a new tool in the registry.
+        with_tool(tool: Tool, *, overwrite: bool = False) -> None:
+            Inserts a new tool.
+        replace_tool(tool: Tool) -> None:
+            Replaces a tool with a new tool.
+            NB. This is a shortcut for `with_tool(tool, overwrite=True)`.
         get_tool(tool_id: str) -> Tool:
             Retrieves a tool by its ID.
         get_tools() -> list[Tool]:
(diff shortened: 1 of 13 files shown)


portia/tool_registry.py
```diff
@@ -342,51 +244,46 @@ class PortiaToolRegistry(ToolRegistry):
...
+            super().__init__(self._load_tools(client))
+        elif config is not None:
+            client = PortiaCloudClient().get_client(config)
+            super().__init__(self._load_tools(client))
+        else:
+            raise ValueError("Either config or client must be provided")
+
+    def with_default_tool_filter(self) -> PortiaToolRegistry:
         """Create a PortiaToolRegistry with a default tool filter."""
-        return PortiaToolRegistry(config).filter_tools(cls.default_tool_filter)
```
robbie-portia, line 366: I don't think we should reduce the `filter_tools()` API - I think that's something that clients may want to do.

tests/integration/test_portia_cloud.py
```diff
@@ -110,14 +108,19 @@ def test_default_portia_has_correct_tools() -> None:
...
     """Test that the default portia has the correct tools."""
-
-    # Choose to exclude gmail rather than microsoft outlook
-    def exclude_gmail_filter(tool: Tool) -> bool:
-        return "gmail" not in tool.id
-
-    registry = PortiaToolRegistry(config=Config.from_default()).filter_tools(exclude_gmail_filter)
-    portia = Portia(tools=registry)
-    tools = portia.tool_registry.get_tools()
+    with patch.object(
```
robbie-portia, line 111: Linked to the above comment, I think we should be able to use Microsoft tools as an end-user without patching SDK.

robbie-portia (commented):
This is loads nicer - great stuff 🚀 Just the one question on tool filtering

portia/tool_registry.py
sam-portia: Agreed - adding this to base-class

tests/integration/test_portia_cloud.py
sam-portia: That should be fine - you can simply instantiate the `PortiaToolRegistry()` without then calling `with_default_tool_filter`

robbie-portia: In that case, I think that's what we should use in this test

portia/tool_registry.py
```diff
@@ -154,113 +200,34 @@ def __radd__(self, other: ToolRegistry | list[Tool]) -> ToolRegistry:
...
-        )
+        other_registry = other if isinstance(other, ToolRegistry) else ToolRegistry(other)
         self_tools = self.get_tools()
         other_tools = other_registry.get_tools()
-        tool_ids = set()
+        tools = {}
         for tool in [*self_tools, *other_tools]:
-            if tool.id in tool_ids:
+            if tool.id in tools:
                 logger().warning(
```
robbie-portia, line 216: We might want to skip the add in this case.

robbie-portia (approved):
Looks really good - just the one comment on the test

portia/tool_registry.py
sam-portia: The idea here was to try and do the least surprising thing, and because last-item-wins is the behaviour when merging dictionaries, it felt like copying that was the best approach

robbie-portia: Ah sorry, yep, I misread! ignore me :D

PR #283, opened 2025-04-04
Change plans to generate steps with inputs only from prev outputs

OmarEl-Mohandes:
# Description

- Modified the planner to generate steps with inputs from past outputs only.
   - This is best effort, we don't fail the plan if it contains non-prev outputs as inputs (it just won't be included in the context -- since no value for those). 
- Remove value from `Variable` and add few tweaks to planning prompt.
- Update tests and evals (on different PRs on platform).

**Note: This might be  a breaking change.**
  -  Fix: I changed the `forbid` to `ignore` in `Variable`, so it can backwards compatible change.

[Ticket Link](https://linear.app/portialabs/issue/POR-1165/[planning-agent]-experiment-with-the-new-structure-of-replacing-inputs)
## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [x] Breaking change 
- [x] Refactor
- [ ] Documentation update

## Screenshots

## Changelog


portia/execution_agents/context.py
```diff
@@ -62,26 +62,17 @@ def generate_input_context(
...
-        if var.value is not None:
+    for ref in inputs:
+        if ref.name in previous_outputs:
             input_context.extend(
                 [
-                    f"input_name: {var.name}",
-                    f"input_value: {var.value}",
-                    f"input_description: {var.description}",
+                    f"input_name: {ref.name}",
+                    f"input_value: {previous_outputs[ref.name].value}",
```
sam-portia, line 70: I thought you were getting rid of the `value` here?

portia/templates/example_plans.py
```diff
@@ -3,39 +3,6 @@
...
-                        name="$email",
-                        value="hello@portialabs.ai",
-                        description="The email address to send the email to",
-                    ),
-                ],
-                tool_id="portia::google_gmail::send_email_tool",
-                output="$final_output",
-            ),
-        ],
-    ),
```
sam-portia, line 38: Why remove this one?

tests/integration/test_e2e.py
```diff
@@ -184,7 +170,7 @@ def test_portia_run_query_with_clarifications_no_handler() -> None:
     config = Config.from_default(
         default_log_level=LogLevel.DEBUG,
         llm_provider=LLMProvider.OPENAI,
-        llm_model_name=LLMModel.GPT_4_O_MINI,
+        llm_model_name=LLMModel.O_3_MINI,
```
sam-portia, line 173: Should we be using `o3-mini` in integrations tests?

portia/execution_agents/context.py
OmarEl-Mohandes: This when representing the input int the context. We need to write the input's value (from previous output).

tests/integration/test_e2e.py
OmarEl-Mohandes: Just wanted to test the default planning model (o3-mini) we're using in one of the cases.

portia/templates/example_plans.py
OmarEl-Mohandes: Didn't see a big value for this one and don't want to overload it with examples. I tested it couple of times and figured it gets it correct everytime.

tests/integration/test_e2e.py
sam-portia: I wonder if we should leave that for the evals? Its only one test so maybe not a huge deal, but o3-mini is ~7x as expensive

sam-portia (approved):

tests/integration/test_e2e.py
OmarEl-Mohandes: good point, will revert it back.

robbie-portia (approved):
LGTM (based on eval results - trusting @sam-portia on the code :D)

PR #292, opened 2025-04-08
[Breaking Changes] Refactor Model configuration settings

sam-portia:
# Description

Essentially this refactor had 3 goals:

1. **LLMProvider is not always needed**: If a user is providing their own custom models, a provider is not necessary. We should not error in this case
2. **Users can pass models as strings**: Industry-standard style: `openai/gpt-4o`, `anthropic/claude-3-7-sonnet-latest` etc.
3. **Users can pass GenerativeModel objects directly**: This way API keys are not required, because they are part of the model object itself.

[This](https://github.com/portiaAI/portia-agent-examples/pull/26/files#diff-b4b06281ae0cbf9a81d41a56114e4837d14154f88033eded57c023e7fdb35169R7-R10) is a good example of what these changes achieve.

*Most* of the changes are in `config.py` and `test_config.py`

`LLMModel` is considered deprecated - it has no use with this refactor and with the removal of `LLMWrapper`
Also removed `LLMWrapper`.

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [x] Breaking change 
- [x] Refactor
- [ ] Documentation update

## Screenshots

## Changelog


portia/model.py
```diff
@@ -67,13 +68,33 @@ def to_langchain(self) -> BaseMessage:
...
+        AZURE_OPENAI: Azure OpenAI provider.
+
+    """
+
+    OPENAI = "openai"
+    ANTHROPIC = "anthropic"
+    MISTRALAI = "mistral"
+    GOOGLE_GENERATIVE_AI = "google"
+    AZURE_OPENAI = "azure-openai"
+    CUSTOM = "custom"
```
sam-portia, line 88: Moved from `config.py`

tests/integration/test_e2e.py
```diff
@@ -33,11 +32,11 @@
 CORE_MODELS = [
     (
         LLMProvider.OPENAI,
-        LLMModel.GPT_4_O_MINI,
+        "openai/gpt-4o-mini",
     ),
     (
         LLMProvider.ANTHROPIC,
-        LLMModel.CLAUDE_3_OPUS,
+        "anthropic/claude-3-5-sonnet-latest",
```
sam-portia, line 39: Changed from Opus to Sonnet 3.5

tests/unit/test_portia.py
```diff
@@ -541,7 +541,7 @@ def test_portia_get_final_output_handles_summary_error(portia: Portia) -> None:
     mock_agent.create_summary.side_effect = Exception("Summary failed")
 
     with mock.patch(
-        "portia.execution_agents.utils.final_output_summarizer.FinalOutputSummarizer",
+        "portia.portia.FinalOutputSummarizer",
```
sam-portia, line 544: Fix broken mock 

Because we do `from portia.execution_agents.utils.final_output_summarizer import FinalOutputSummarizer` in `portia.portia` - the name to mock has to be `portia.portia.FinalOutputSummarizer`.

If we did `import portia.execution_agents.utils.final_output_summarizer.FinalOutputSummarizer` or `from portia.execution_agents.utils.final_output_summarizer import FinalOutputSummarizer` then the previous mock would be fine

portia/config.py
```diff
@@ -407,20 +386,36 @@ class Config(BaseModel):
         description="The endpoint for Azure OpenAI. Must be set if llm-provider is AZURE_OPENAI",
     )
 
-    llm_provider: LLMProvider = Field(
-        default=LLMProvider.OPENAI,
-        description="Which LLM Provider to use.",
+    llm_provider: LLMProvider | None = Field(
+        default=None,
+        description="Which LLM Provider to use for resolving GenerativeModels. Can be None if "
```
emmaportia, line 391: I think it would be helpful if we could explain to people that if they specify the llm_provider, we'll automatically select the right models within that for high performance / cost, and then explain what the interaction is vs models.

portia/config.py
```diff
@@ -405,20 +386,36 @@ class Config(BaseModel):
...
+        "GenerativeModel instances are provided directly.",
     )
 
-    models: dict[str, LLMModel] = Field(
+    models: dict[str, str | GenerativeModel] = Field(
```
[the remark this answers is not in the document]
emmaportia: I think this structure of dict[str, str | GenerativeModel] is extremely hard to understand. I know this code and I find it hard to understand! We don't really document or explain what is supposed to be the key or how it works at the moment.

portia/config.py
emmaportia: can we also update the documentation on line 332?

portia/config.py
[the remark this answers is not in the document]
sam-portia: Just to be clear - are you saying its worse than `dict[str, LLMModel]` or that it was always a bit confusing?

The `str | GenerativeModel` I've found to have pretty good ergonomics when it comes to using it fwiw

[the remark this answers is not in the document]
emmaportia: no no, its always been bad, i was just hoping this would make it clearer :)

tests/integration/test_e2e.py
```diff
@@ -33,11 +32,11 @@
 CORE_MODELS = [
     (
         LLMProvider.OPENAI,
-        LLMModel.GPT_4_O_MINI,
+        "openai/gpt-4o-mini",
```
emmaportia, line 35: we're using 4o-mini as the default?

portia/config.py
```diff
@@ -316,21 +281,34 @@ def parse_str_to_enum(value: str | E, enum_type: type[E]) -> E:
...
+class GenerativeModels(BaseModel):
+    """Configuration for a GenerativeModels."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    default_model: GenerativeModel | str | None = None
+    planning_model: GenerativeModel | str | None = None
+    execution_model: GenerativeModel | str | None = None
+    introspection_model: GenerativeModel | str | None = None
+    summariser_model: GenerativeModel | str | None = None
```
sam-portia, line 311: Change 1: Make the model configuration explicit and structured

portia/config.py
```diff
@@ -684,9 +633,117 @@ def must_get(self, name: str, expected_type: type[T]) -> T:
...
+            # Default model is required, but not provided.
+            raise InvalidConfigError(
+                "default_model",
+                "A default model must be set",
+            )
+        return self._build_generative_model(model)
+
+    def get_planning_model(self) -> GenerativeModel:
+        """Get or build the planning model from the config."""
+        return self._build_generative_model(self.models.planning_model) or self.get_default_model()
```
sam-portia, line 650: Change 2: Explicit `get_` methods for the models that can be configured

sam-portia (commented):
@emmaportia 

Given the config a rethink - see the 2 changes I highlighted. 

This removes the ability for a user to register their own usage, but realistically I dont see why they would be doing that - if they define their own agent or tool they can either use one of the existing models or instantiate their own model for the purpose.

portia/config.py
sam-portia: We can add docs against each one of these fields

portia/common.py
```diff
@@ -57,6 +57,7 @@ def combine_args_kwargs(*args: Any, **kwargs: Any) -> Any:  # noqa: ANN401
 
 EXTRAS_GROUPS_DEPENDENCIES = {
     "mistral": ["mistralai", "langchain_mistralai"],
+    "mistralai": ["mistralai", "langchain_mistralai"],  # in process of refactoring
```
emmaportia, line 60: pray tell...?

portia/portia.py
```diff
@@ -853,9 +853,14 @@ def _log_replan_with_portia_cloud_tools(
         example_plans: list[Plan] | None = None,
     ) -> None:
         """Generate a plan using Portia cloud tools for users who's plans fail without them."""
-        cloud_registry = self.tool_registry + PortiaToolRegistry.with_unauthenticated_client(
+        unauthenticated_client = PortiaCloudClient.new_client(
             self.config,
+            allow_unauthenticated=True,
```
emmaportia, line 858: what does allow_unauthenticated do?

portia/config.py
emmaportia: Yeah I think this is better. @TomSPortia since he created a bunch of the config stuff originally

TomSPortia: Agreed, as we're leaning into a set of pre-baked agents I think we should be explicit and if users have other use-cases for models they can manage that themselves.

portia/common.py
sam-portia: I want to make our naming of mistral as a provider consistent - we mix and match between mistral and mistralai and it felt sloppy (mainly my fault). So adding the new before removing the old

portia/portia.py
sam-portia: It means the requests don’t have the auth header automatically added, which we need in this case to fetch tools as an unauthenticated user. 

We do actually have a ticket to tidy up the client - we don’t really need to be recreating the http client just for one request with different headers

tests/integration/test_e2e.py
sam-portia: Just for these integration tests.. although perhaps this var name is misleading, it should really be `CORE_PROVIDERS`

emmaportia (approved):

PR #295, opened 2025-04-09
Ollama support

sam-portia:
# Description

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/config.py
+++ b/portia/config.py
@@ -432,6 +432,10 @@ class Config(BaseModel):
         default_factory=lambda: os.getenv("AZURE_OPENAI_ENDPOINT") or "",
         description="The endpoint for Azure OpenAI. Must be set if llm-provider is AZURE_OPENAI",
     )
+    ollama_base_url: str = Field(
+        default_factory=lambda: os.getenv("OLLAMA_BASE_URL") or "http://localhost:11434/v1",
+        description="The base URL for Ollama. Must be set if llm-provider is OLLAMA",
+    )
 
     llm_provider: LLMProvider | None = Field(
         default=None,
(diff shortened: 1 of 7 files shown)


portia/config.py
```diff
@@ -432,6 +432,10 @@ class Config(BaseModel):
         default_factory=lambda: os.getenv("AZURE_OPENAI_ENDPOINT") or "",
         description="The endpoint for Azure OpenAI. Must be set if llm-provider is AZURE_OPENAI",
     )
+    ollama_base_url: str = Field(
```
Nathanjp91, line 435: can we just make this api_url: str | None, ollama isn't the only local llm caller, and even for things like OpenAI we might want to overwrite the api_url to support openai specs that are hosted on AWS or via proxy systems.

Nathanjp91: None by default also feels cleaner than having it set by default imo

tests/integration/test_model.py
```diff
@@ -101,9 +109,13 @@ def test_get_structured_response(model_str: str, messages: list[Message]) -> Non
     assert response.message is not None
 
 
-@pytest.mark.parametrize("model_str", MODELS)
+@pytest.mark.parametrize("model_str", [M for M in MODELS if not M.startswith("ollama/")])
```
Nathanjp91, line 112: confused why you add it above just to ignore it here?

sam-portia: Its used in the other tests that are parametrized by MODELS, just not this one

portia/config.py
sam-portia: > can we just make this api_url: str | None, ollama isn't the only local llm caller, and even for things like OpenAI we might want to overwrite the api_url to support openai specs that are hosted on AWS or via proxy systems.

This I dont think we can do, because a user might use two different providers that both use an `api_url`. I do have a to-do to clean up api-keys and provider-related config too

sam-portia: > None by default also feels cleaner than having it set by default imo

Happy to do this, I was mostly copying the convention we had so far where pretty much everything is not nullable

sam-portia: Actually I kind of want to leave the Ollama one with a default because it runs locally on a standard port, I dont think we should force the user to always specify it

tests/integration/test_model.py
sam-portia: Cleared this up

Nathanjp91 (approved):

PR #304, opened 2025-04-10
User Led Learning: SDK similarity search

Nathanjp91:
# Description

Adds in function on portia to do similarity search on query with optional limit/threshold for the similarity endpoint. Defaults provided. 

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/portia.py
+++ b/portia/portia.py
@@ -28,6 +28,7 @@ from portia.clarification import (
     Clarification,
     ClarificationCategory,
 )
+from portia.cloud import PortiaCloudClient
 from portia.config import (
     Config,
     ExecutionAgentType,
@@ -66,7 +67,6 @@ from portia.storage import (
 from portia.tool import ToolRunContext
 from portia.tool_registry import (
     DefaultToolRegistry,
-    InMemoryToolRegistry,
     PortiaToolRegistry,
     ToolRegistry,
 )
@@ -127,7 +127,7 @@ class Portia:
         if isinstance(tools, ToolRegistry):
             self.tool_registry = tools
         elif isinstance(tools, list):
-            self.tool_registry = InMemoryToolRegistry.from_local_tools(tools)
+            self.tool_registry = ToolRegistry(tools)
         else:
             self.tool_registry = DefaultToolRegistry(self.config)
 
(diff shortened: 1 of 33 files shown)


portia/portia.py
```diff
@@ -897,3 +899,35 @@ def _persist_step_state(self, plan_run: PlanRun, step: Step) -> None:
             plan_run.outputs.step_outputs[step.output] = step_output
 
         self.storage.save_plan_run(plan_run)
+
+    def similar_plans(self, query: str, threshold: float = 0.5, limit: int = 5) -> list[Plan]:
```
Nathanjp91, line 903: threshold might seem low here, but cosine distance (which the vector embeddings uses) is a long tail distribution.

TomSPortia, line 903: Looking at the PR in reality I feel like this would fit better on the `PlanStorage` interface. It's similar to the `match_tools` function on the tool registry. We could have the disk + memory ones return [] or self.plans and then properly implement it on the cloud one. That allows people to implement it themselves if they'd like

Nathanjp91: Not disagreeing that the retrieval code feels more natural on the storage object, but 2 things to note here I think.

1. Tried extending this with a custom storage class, and if we want people to be able to extend we probably have to refactor our storage class code becasue atm you actually can't supply a custom storage class outside of our tightly coupled Enum in practice.  Anytime we use enums in our code it comes at the cost of extensibility. We should use them sparingly and only on things that are not expected to be customized.

2. It does make the flow a bit more awkward in terms of whole process implementation
```
examples = portia.storage.get_similar_plans(query)
plan = portia.plan(query, examples)
```
Dips into a sub field of portia to then call a top level function, feels like I'm accessing internals. 

What's your thoughts on 2? do we extend portia.plan to retrieve examples automatically if it's got api key? Extend function with the fields that similar plans requires? Then we're adding a bunch of function headers that might not be used in local running mode. 

I feel like calling get_similar_plans and just using default settings is fine inside .plan(...), then if they want to do something with threshold and limits they can call portia.storage.get_similar_plans and provide examples directly if they want something more complicated.

TomSPortia (approved):

PR #312, opened 2025-04-14
Fix two flaky tests

OmarEl-Mohandes:
# Description

``` Test: test_portia_run_query_requiring_cloud_tools_not_authenticated ``` is failing 1 in 5, duo to producing plans with 0 steps.

 - Fix: Made some planning prompt edits to highlight it should return error instead.

``` Test: test_portia_run_query_with_example_registry ``` is failing 1 in 7, duo haiku step is selecting search_tool instead of llm_tool. 

 - Fix: Made some minor edit in the query to nudge it towards LLM tool.

Both now are passing 10 times in a row, so should be more stable.

## Type of change

(select all that apply)

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/planning_agents/default_planning_agent.py
+++ b/portia/planning_agents/default_planning_agent.py
@@ -53,8 +53,8 @@ IMPORTANT GUIDLINES:
 - When creating the description for a step of the plan, if you need information from the previous
  step, DO NOT guess what that step will produce - instead, specify the previous step's output as an
  input for this step and allow this to be handled when we execute the plan.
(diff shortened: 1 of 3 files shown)


OmarEl-Mohandes:
I can double check but I don't think so. I can put an item for it.

OmarEl-Mohandes:
> 👌 thanks!
> 
> Do we have an eval for a query that has no matching tools?


I can double check but I don't think so. I can put an item for it.

PR #313, opened 2025-04-14
Expose Config.get_model, rename GenerativeModelsConfig and expose more types from __init__

sam-portia:
# Description

* `get_model` exposed allows us to instantiate models from a canonical string like `openai/gpt-4o` which makes configuring Tool models more ergonomic
* Renamed `GenerativeModels` to `GenerativeModelsConfig` 
* Exposed main `Model` types from `portia.__init__`

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [x] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/config.py
+++ b/portia/config.py
@@ -306,8 +306,8 @@ PROVIDER_DEFAULT_MODELS = {
 }
 
 
-class GenerativeModels(BaseModel):
-    """Configuration for a GenerativeModels.
+class GenerativeModelsConfig(BaseModel):
+    """Configuration for a Generative Models.
 
     These models do not all need to be specified manually. If an LLM provider is configured,
     Portia will use default models that are selected for the particular use-case.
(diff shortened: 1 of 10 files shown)


portia/config.py
```diff
@@ -688,47 +688,46 @@ def get_default_model(self) -> GenerativeModel:
...
+        See the GenerativeModelsConfig class for more information
         """
-        return (
-            self._construct_generative_model(self.models.summarizer_model)
-            or self.get_default_model()
-        )
+        return self.get_model(self.models.summarizer_model) or self.get_default_model()
 
-    def _construct_generative_model(
+    def get_model(
```
OmarEl-Mohandes, line 716: Have to say, I'm a bit confused with `get_model` that that takes the `model` and returns it back, except if it's a string. Do you think `get_generative_model` would be more illustrative? .. it's a bit verbose maybe.

OmarEl-Mohandes (approved):

portia/config.py
sam-portia: No I agree that's better

PR #316, opened 2025-04-14
Fix cli.py

robbie-portia:
# Description

Without this, I get the below error with cli.py
```
poetry run python -m portia.cli run "Get the weather in London, then write a poem about Robbie Heywood in London experiencing the current weather, then email it to robbie@portialabs.ai" --env-location ENV_FILE --default-log-level debug
2025-04-14 20:01:30.387 | ERROR | __main__:_get_config:352 - Config value NONE is not valid - Value must be a string or LLMProvider
```

## Type of change

(select all that apply)

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/cli.py
+++ b/portia/cli.py
@@ -141,6 +141,8 @@ def generate_cli_option_from_pydantic_field(  # noqa: C901, PLR0912
                     field_default = info.default_factory().name  # type: ignore[reportCallIssue]
                 else:
                     field_default = None
+            else:
+                return lambda: None
 
     field_help = info.description or f"Set the value for {option_name}"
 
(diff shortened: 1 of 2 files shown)


sam-portia (approved):

robbie-portia:
@sam-portia - I've just made an update - it now actually fixes CLI (sorted original fix and also had to make it ignore other fields so that it ignores the `models` field - this means there is no way to set that through the CLI for now).

sam-portia (approved):
Thanks - I will make sure the models stuff works in a consistent way, I hadn’t really got around to that

PR #319, opened 2025-04-15
Execution agent uses memory

robbie-portia:
# Description

Adds the ability for the execution agent to pull previous outputs from agent memory for use as inputs in a future step.

Implementation details:
* A new 'memory extraction step' is added as the first step of the execution graph in langgraph
* It pulls any of the inputs out of agent memory if they are in agent memory and builds a list of step_inputs
* These step inputs are then used in the parser / verifier input when building the context

The design is a little different to the one we discussed previously - I chatted this through with @emmaportia yesterday. Previously, we discussed that the 'memory extraction step' would use an LLM to only pull out input values that might need some extraction (e.g. if we had a list of slots used for a meeting and we need to choose an empty slot) and otherwise the parser / verifier would use a template for the large output values, with the then being templated in after the tool agent. I do have the code for this, but it was getting a little complex and I worried it would be brittle (it was hard to get it to reliably pick out values for extraction). This design is significantly simpler, at the expense of us not handling inputs that are larger than the size of the context window (because the large inputs will still need to go through the parser / verifier / tool agent). Given the size of current context windows (and the fact they are increasing), I think that's an ok trade-off for now, but let me know if you disagree.

## Type of change

(select all that apply)

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update


portia/execution_agents/default_execution_agent.py
```diff
@@ -168,6 +238,9 @@ class ParserModel:
                 "- You may take values from the task, inputs, previous steps or clarifications\n"
                 "- Prefer values clarified in follow-up inputs over initial inputs.\n"
                 "- Do not provide placeholder values (e.g., 'example@example.com').\n"
+                "- Do not include references to any of the input values (e.g. 'as provided in "
+                "the input') - you must put the exact value the tool should be called with in "
```
TomSPortia, line 242: nit: new line between bullet points

portia/execution_agents/one_shot_agent.py
```diff
@@ -174,7 +179,22 @@ def execute_sync(self) -> Output:
         if not self.tool:
             raise InvalidAgentError("No tool available")
 
-        context = self.get_system_context()
+        # Note: the one-shot agent does not support pulling outputs from agent memory
```
TomSPortia, line 182: Should this be an error? Seems like it could lead to weird behaviour and silent failures

portia/execution_agents/default_execution_agent.py
```diff
@@ -117,6 +127,66 @@ class VerifiedToolInputs(BaseModel):
...
+
+        Returns:
+            dict[str, Any]: The LangGraph state update.
+
+        """
+        step_inputs = []
+        previous_outputs = self.agent.plan_run.outputs.step_outputs
+        for step_input in self.agent.step.inputs:
+            if step_input.name not in previous_outputs:
+                logger().warning(
```
TomSPortia, line 159: Should this happen? I think my gut says we should error here as the plan is likely inconsistent (trigger replanning?) but not a strong opinion

portia/execution_agents/default_execution_agent.py
```diff
@@ -117,6 +127,66 @@ class VerifiedToolInputs(BaseModel):
...
+                    "Received step unknown step input with name: %s, skipping...",
+                    step_input.name,
+                )
+                continue
+            previous_output = previous_outputs.get(step_input.name)
+
+            if isinstance(previous_output, LocalOutput):
```
TomSPortia, line 166: nit: match statement here might be more extensible?

TomSPortia (commented):
Looks great, just some small style/error things

portia/execution_agents/one_shot_agent.py
robbie-portia: Good call - I've now made it an error

portia/execution_agents/default_execution_agent.py
robbie-portia: This actually wasn't meant to be a new bullet, but was an extension of the other one. I've changed to use a comma now to make it clearer.

portia/execution_agents/default_execution_agent.py
robbie-portia: Good call - done

portia/execution_agents/default_execution_agent.py
robbie-portia: Yeah, I wasn't sure on this one with whether we should try to soldier on or call out the issue. I think you're probably write that soldiering on with a potentially inconsistent plan isn't right, so I've now made this an error.

TomSPortia (approved):

PR #324, opened 2025-04-16
Update end user handling

TomSPortia:
# Description

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog


example.py
```diff
@@ -21,18 +22,21 @@
...
-# We can also provide additional execution context to the process
-with execution_context(end_user_id="123", additional_data={"email_address": "hello@portialabs.ai"}):
-    plan_run = portia.run(
-        "Get the temperature in London and Sydney and then add the two temperatures rounded to 2DP",
-    )
+# We can also provide additional data about the end user to the process
+
+plan_run = portia.run(
+    "Please email me with a haiku about the weather in london",
+    end_user=EndUser(
```
TomSPortia, line 29: The function signature here is end_user: `EndUser | str | None`

allowing you to:
- pass in nothing and we default to the end user of the API key owner
- pass in a string like today and we will create the end user model
- pass in a full end user model and we will create it along with any attributes you pass.

portia/execution_agents/context.py
```diff
@@ -166,17 +167,20 @@ def generate_context_from_execution_context(context: ExecutionContext) -> list[s
...
-
     execution_context = ["Metadata: This section contains general context about this execution."]
-    if context.end_user_id:
-        execution_context.extend(
-            [
-                f"end_user_id: {context.end_user_id}",
-            ],
-        )
-    for key, value in context.additional_data.items():
+    execution_context.extend(
```
TomSPortia, line 171: I have a branch with some evals for this

example.py
```diff
@@ -21,18 +22,21 @@
...
+        external_id="user_789",
+        email="tom@portialabs.ai",
```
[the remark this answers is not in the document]
OmarEl-Mohandes: Or you might you actually need to change the whole example, because it depends on email tool and i think the example should only work on `example_tool_registry`.

portia/plan_run.py
```diff
@@ -112,6 +112,10 @@ class PlanRun(BaseModel):
         default=empty_context(),
         description="Execution Context for the PlanRun.",
     )
+    end_user_id: str = Field(
```
OmarEl-Mohandes, line 115: Should this be id only or EndUser model?

portia/portia.py
```diff
@@ -140,11 +141,20 @@ def __init__(
             case StorageClass.CLOUD:
                 self.storage = PortiaCloudStorage(config=self.config)
 
+    def handle_end_user(self, end_user: str | EndUser | None = None) -> EndUser:
```
OmarEl-Mohandes, line 144: I'd prefer more concrete name? e.g. `get_or_save`.

Also, to confirm this explicitly saves once and gets otherwise. So it shouldn't save end user if it's changed?

portia/portia.py
```diff
@@ -156,20 +166,23 @@ def run(
...
+            end_user (str | EndUser | None = None): The end user for this plan run.
 
         Returns:
             PlanRun: The run resulting from executing the query.
 
         """
-        plan = self.plan(query, tools, example_plans)
-        plan_run = self.create_plan_run(plan)
+        plan = self.plan(query, tools, example_plans, end_user)
+        end_user = self.handle_end_user(end_user)
```
OmarEl-Mohandes, line 176: Wanna confirm from backend perspective, no dependency between end_user model and plan? so any order should be fine?

portia/storage.py
```diff
@@ -446,8 +469,34 @@ def get_plan_run_output(self, output_name: str, plan_run_id: PlanRunUUID) -> Loc
...
+
+        Args:
+            end_user (EndUser): The EndUser object to save.
+
+        """
+        self.end_users[end_user.external_id] = end_user
+        return end_user
 
-class DiskFileStorage(PlanStorage, RunStorage, LogAdditionalStorage, AgentMemory):
+    def get_end_user(self, external_id: str) -> EndUser:
```
OmarEl-Mohandes, line 486: I wonder if `get_end_user` should just gets or None rather than saves as well. Otherwise, from user pov no way they can check if end_user just exists in storage without saving?

I presume this behaviour applies to all different storages.

portia/execution_agents/context.py
OmarEl-Mohandes: Thanks! .. I think you'll need to provide more details into the planning prompt on what to do this this new info.

OmarEl-Mohandes (commented):
Thanks! .. Mostly minor comments and some suggestions/questions.

example.py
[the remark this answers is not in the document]
TomSPortia: good spot!

portia/cli.py
```diff
@@ -189,16 +188,15 @@ def run(
         execution_hooks=CLIExecutionHooks(),
     )
 
-    with execution_context(end_user_id=cli_config.end_user_id):
-        plan = portia.plan(query)
+    plan = portia.plan(query, end_user=cli_config.end_user_id)
```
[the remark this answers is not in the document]
TomSPortia: you don't need it, it follows the same semantics as run, i.e if not provided get the default

portia/end_user.py
```diff
@@ -0,0 +1,30 @@
...
+class EndUser(BaseModel):
+    """Represents an actual user of the system."""
+
+    external_id: str = Field(description="The external ID of the end user.")
+
+    name: str = Field(default="", description="The name of the end user.")
+    email: str = Field(default="", description="The email address of the end user.")
+    phone_number: str = Field(default="", description="The phone number of the end user.")
+
+    additional_data: dict[str, str] = Field(
```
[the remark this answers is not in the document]
TomSPortia: I'm against this because it leads to lots of weird bugs when the value can't be serialized to a string. Enforcing it as [str, str] means the user has to serialize it which for now I  prefer. If we get lots of feedback on this we can look again.

portia/portia.py
TomSPortia: there is a linkage so the end_user needs to exist, but we handle this initialisation up front.

portia/storage.py
TomSPortia: updated!

OmarEl-Mohandes (approved):
LGTM 🚀

PR #329, opened 2025-04-17
Introspection agent uses memory

robbie-portia:
# Description

Introspection agent pulls in values from agent memory when required.

I also realised I hadn't added the output max size, so including that too.

## Type of change

(select all that apply)

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog


portia/introspection_agents/default_introspection_agent.py
```diff
@@ -79,6 +85,20 @@ def pre_step_introspection(
         plan_run: PlanRun,
     ) -> PreStepIntrospection:
         """Ask the LLM whether to continue, skip or fail the plan_run."""
+        introspection_condition = plan.steps[plan_run.current_step_index].condition
+        if not introspection_condition:
+            return PreStepIntrospection(
+                outcome=PreStepIntrospectionOutcome.CONTINUE,
+                reason="No condition provided for step",
+            )
```
OmarEl-Mohandes, line 93: This is not needed. It's already handled on [the caller side](https://github.com/portiaAI/portia-sdk-python/blob/b4b60b4346e20d6fc10663241f8c8538f81531ac/portia/portia.py#L672-L681). Happy to do it here or there, but let's make sure to update tests if it's here.

portia/introspection_agents/default_introspection_agent.py
```diff
@@ -79,6 +85,20 @@ def pre_step_introspection(
...
+
+        memory_outputs = [
+            self.agent_memory.get_plan_run_output(output.output_name, plan_run.id)
+            for output in plan_run.outputs.step_outputs.values()
+            if isinstance(output, AgentMemoryOutput)
+            and output.output_name in introspection_condition
```
OmarEl-Mohandes, line 99: We might need to include all outputs regardless, not just in the condition because of COMPLETE/FAIL outcomes.

OmarEl-Mohandes (commented):
Thanks! .. just minor comments.

portia/introspection_agents/default_introspection_agent.py
robbie-portia: Makes sense - I needed it to avoid a type error, but I've just added it as a check in the list comprehension now.

portia/introspection_agents/default_introspection_agent.py
robbie-portia: I don't think we do want to pull in all the outputs - otherwise there isn't really much point in doing the agent memory because we're pulling everything in anyway. We'll still have summaries for everything - I think it's reasonable to only pull in big values if they are actually used.

PR #333, opened 2025-04-20
Action clarifications with confirmation

emmaportia:
# Description

Browser tools require the user to take action and then confirm so wanted to add a way to do this which is backwards compatible.

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [X] Requires sync with platform release
- [ ] Documentation update

## Screenshots

### Added

- vTODO. Added a new type of action clarification that requires the user to indicate the action has been completed.


emmaportia:
@TomSPortia -- mind taking a look at this directionally before i round out the changes and make the associated backend changes?

TomSPortia (commented):
It feels a little off to me to have a new category of clarifications just for this. Could we not just extend the ActionClarification model to be:

```python
class ActionClarification(Clarification):
   require_confirmation: bool = Field(
        default=False,
        description="Whether human confirmation is required.",
    )

```

and then most of the code doesn't need to change as the default keeps it backwards compatible?

PR #337, opened 2025-04-22
Add extra argument to LLMTool `input_data: list[str]`

sam-portia:
# Description

The Execution Agent sometimes fails to pass inputs to the LLM Tool - it seems to think that the `task` argument should be the same as the Step task parameter, but that then misses inputs (variables and other context).

See doc for context of the decisions that went into fixing this: https://docs.google.com/document/d/1p3I8M9YnL5w2o1JMmFevbGrCHqWH0b0PiLY_bpSuTJE/edit?tab=t.0

Ticket Link: N/A 

## Type of change

(select all that apply)

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/open_source_tools/llm_tool.py
+++ b/portia/open_source_tools/llm_tool.py
@@ -22,8 +22,9 @@ class LLMToolSchema(BaseModel):
     )
     input_data: list[str] = Field(
         default_factory=list,
-        description="Any relevant data that should be used to complete the task. This should "
-        "include all relevant data in their entirety (i.e. not a summary).",
+        description="Any relevant data that should be used to complete the task. Important: This "
+        "should include all relevant data in their entirety, from the first to the last character "
+        "(i.e. NOT a summary).",
     )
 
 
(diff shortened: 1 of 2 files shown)


portia/open_source_tools/llm_tool.py
```diff
@@ -60,7 +69,7 @@ class LLMTool(Tool[str]):
         "the model will be resolved from the config.",
     )
 
-    def run(self, ctx: ToolRunContext, task: str) -> str:
+    def run(self, ctx: ToolRunContext, task: str, input_data: Sequence[str] = ()) -> str:
```
Nathanjp91, line 72: Does using sequence avoid the global function default issue?

Nathanjp91 (approved):
LGTM: Provided that global issue isn't a problem

portia/open_source_tools/llm_tool.py
sam-portia: Using a tuple does yes.. The Sequence type is to make sure a `list` is also valid

PR #347, opened 2025-04-23
Anthropic models: Increase max tokens and fall back on instructor for structured output

sam-portia:
# Description

Claude 3.5 defaults to 1024 tokens output which breaks the execution agent for larger (tool) inputs. This is a problem for the LLM tool, which sometimes need passing the large values via the new `input_data: list[str]` arg.

Claude 3.5 and 3.7 also struggle with structuring the JSON for large inputs into tool arguments (see eval PR link below). 

Instructor JSON mode doesnt have this problem, but it does introduce overhead of extra input tokens and introduces a separate problem where Claude writes a message before the JSON (because we are using Instructor `Mode.ANTHROPIC_JSON`)

For this reason, I've implemented the instructor mode as a fall back ONLY when Claude is trying to parse a large value into a JSON (which you can determine by asking Langchain to return raw responses)

Related eval PR https://github.com/portiaAI/platform/pull/384

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/model.py
+++ b/portia/model.py
@@ -461,18 +461,17 @@ class AnthropicGenerativeModel(LangChainGenerativeModel):
             return self.get_structured_response_instructor(messages, schema)
         langchain_messages = [msg.to_langchain() for msg in messages]
         structured_client = self._client.with_structured_output(schema, include_raw=True, **kwargs)
-        response = structured_client.invoke(langchain_messages)
+        raw_response = structured_client.invoke(langchain_messages)
+        if not isinstance(raw_response, dict):
+            raise TypeError(f"Expected dict, got {type(raw_response)}")
         # Anthropic sometimes struggles serializing large JSON responses, so we fall back to
         # instructor if the response is above a certain size.
-        if (
-            isinstance(response, dict)
-            and (response.get("parsing_error") or {}).get("error") == "ValidationError"
-            and (
-                len(tiktoken.get_encoding("gpt2").encode(json.dumps(response["raw"])))
-                > self._output_instructor_threshold
-            )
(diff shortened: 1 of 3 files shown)


portia/model.py
```diff
@@ -450,7 +458,22 @@ def get_structured_response(
...
+        raw_response = structured_client.invoke(langchain_messages)
+        if not isinstance(raw_response, dict):
+            raise TypeError(f"Expected dict, got {type(raw_response)}")
+        # Anthropic sometimes struggles serializing large JSON responses, so we fall back to
+        # instructor if the response is above a certain size.
+        if isinstance(raw_response.get("parsing_error"), ValidationError) and (
+            len(tiktoken.get_encoding("gpt2").encode(raw_response["raw"].model_dump_json()))
+            > self._output_instructor_threshold
+        ):
+            return self.get_structured_response_instructor(messages, schema)
```
robbie-portia, line 472: How does instructor avoid this issue? I might be misunderstanding, but surely it's using the anthropic client under the hood?

sam-portia: Its using plain-old chat completion rather than structured (tool-calling) mode with its own retry loop

Instructor can use the tool-calling mode too (ANTHROPIC_TOOLS) but then this has the same problem as `with_structured_output`

robbie-portia: Ah ok, makes sense

robbie-portia (approved):
LGTM

PR #355, opened 2025-04-25
Add plan inputs

robbie-portia:
# Description

Vibe coded this just to show how I was imagining it being done - if other people think this looks reasonable, I think it would v quick to touch up a few bits, vibe code some tests and then give it some testing. In theory, it shouldn't really change our performance at all if you don't specify them, so should hopefully all be benefit.

## Type of change

(select all that apply)

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog


.cursor/rules
```diff
@@ -0,0 +1,21 @@
+# Portia: Multi-Agent System Framework
```
robbie-portia, line 1: I thought this would be a useful starting point for iterating on something like this

tests/unit/execution_agents/test_default_execution_agent.py
```diff
@@ -941,18 +941,19 @@ def test_memory_extraction_step_errors_with_missing_input() -> None:
         memory_extraction_step.invoke({"messages": [], "step_inputs": []})
 
 
-def test_memory_extraction_step_handles_unknown_output_type() -> None:
```
robbie-portia, line 944: We don't need this test any more as I updated the way we handle outputs in the memory extraction step

robbie-portia:
I've now sorted the tests and done a little bit of manual testing on this and I think it's now ready for a real review. In summary:
* Plans now have inputs, which you can pass in when planning. These have a name, description and an optional schema for the value.
* The planner is then able to use these as inputs to steps in the same was it uses previous step outputs.
  * The same is true for the PlanBuilder
* When running the plan, you have to specify a value for each plan input.
* The execution agent is then able to use these plan input values as inputs for steps in the same way that it uses previous step outputs.

Currently, as we don't store the plan inputs, this only works if using local / disk storage. If we're happy with this approach, I'll follow up with: adding them in the frontend, adding them into storage, adding a couple of evals and updating the docs.

Let me know what you think :)

portia/execution_agents/default_execution_agent.py
```diff
@@ -155,33 +153,29 @@ def invoke(self, _: ExecutionState) -> dict[str, Any]:
...
-                        "Received unknown output type: %s",
-                        previous_output,
-                    )
-                    continue
+        plan_inputs = self.agent.plan_run.plan_inputs
+
+        for input_variable in self.agent.step.inputs:
+            input_value = None
+
+            if input_variable.name in previous_outputs:
```
emmaportia, line 161: what happened to the agent memory stuff?

portia/execution_agents/default_execution_agent.py
```diff
@@ -155,33 +153,29 @@ def invoke(self, _: ExecutionState) -> dict[str, Any]:
         """
         step_inputs = []
         previous_outputs = self.agent.plan_run.outputs.step_outputs
```
emmaportia, line 155: nit: is it not neater and simpler to just append plan_inputs here? (with some naming changes)

portia/plan.py
```diff
@@ -197,6 +223,63 @@ def pretty_print(self) -> str:
...
+
+    Args:
+        name (str): The name of the input, e.g. $api_key.
+        description (str): A description of the input.
+        value_schema (type[BaseModel] | None): Optional Pydantic model for validating
+            the input value.
+
+    """
+
+    model_config = ConfigDict(extra="ignore", frozen=True)
```
emmaportia, line 237: do you want to set frozen=True here? seems limiting without necessarily a benefit

emmaportia (commented):
This looks good to me generally, a couple of questions:
- Do plan inputs get integrated seamlessly with agent memory (I think so but wanted to check)
- How will you roll it out given it'll cause backend issues if rolled out without backend changes?

portia/execution_agents/default_execution_agent.py
robbie-portia: I think we have to handle them differently because we have to call `.full_value()` for previous outputs because they might be in agent memory (while plan inputs won't be in agent memory), and we also only want to do that for outputs we're actually using, not for all previous outputs. What do you think?

portia/execution_agents/default_execution_agent.py
robbie-portia: The agent memory stuff is still here, I just realised there was an easier way to do it - by using `input_value = previous_output.full_value(self.agent.agent_memory)` we get the behaviour we want in both the local and the memory case. Or was there some other memory stuff you were expecting to see?

robbie-portia:
Thanks Emma - on the two questions:
* I think they work together - essentially, when we're pulling inputs for the step in, we can either take them from previous outputs (local or memory) or take them from plan inputs. We don't store plan inputs in agent memory, but I think that's ok for now - we could add that in future if people need it.
* In terms of rollout, I was thinking that if we're broadly happy with this (which it sounds like we are), I'll quickly knock up the platform changes needed (allow usage in storage + in frontend), merge that first (to enable things in the platform), then merge this so it rolls out smoothly.

emmaportia:
Sounds good to me in terms of approach

portia/execution_agents/default_execution_agent.py
emmaportia: been thinking about this some more -- even if plan inputs won't be in agent memory, it seems plausible in the future that we might want them to go into agent memory so i feel like generally we don't really want the execution agent to have a different mental model of the two of them.

It *might* not be necessary with the current implementation of agent memory, but if you want to e.g template in plan inputs because a plan input is really big, wouldn't you want to be able to do that?

robbie-portia: Thanks for pushing this - I changed LocalOutput -> LocalDataValue in a separate PR and now I've moved PlanInput to taking a LocalDataValue rather than a raw value and I agree it makes things nicer - thanks :D

portia/execution_agents/output.py
```diff
@@ -54,12 +54,12 @@ class LocalDataValue(BaseOutput):
 
     value: Serializable | None = Field(
         default=None,
-        description="The output of the tool.",
+        description="The value.",
```
emmaportia, line 57: nit: "The value, often the output from the tool"?

portia/execution_agents/output.py
```diff
@@ -54,12 +54,12 @@ class LocalDataValue(BaseOutput):
...
     )
 
     summary: str | None = Field(
         default=None,
-        description="Textual summary of the output of the tool. Not all tools generate summaries.",
+        description="Textual summary of the value.",
```
emmaportia, line 62: Still worth putting hte note that not all tools generate summaries (presumably plan inputs wont' have summaries either)?

portia/plan.py
```diff
@@ -197,6 +223,64 @@ def pretty_print(self) -> str:
         return f"{self.name}: ({self.description})"
 
 
+class PlanInput(BaseModel):
+    """An input to a plan.
+
+    Args:
+        name (str): The name of the input, e.g. $api_key.
+        description (str): A description of the input.
+        value_schema (type[BaseModel] | None): Optional Pydantic model for validating
```
emmaportia, line 232: why have this?

emmaportia (commented):
Thanks!

portia/execution_agents/output.py
robbie-portia: Done

portia/execution_agents/output.py
robbie-portia: Done

portia/plan.py
robbie-portia: This was something that @TomSPortia mentioned as a nice-to-have for type checking of input schemas when you run the plan (I think because he's working with some quite nested JSON structured). Happy to remove if you think it isn't needed?

robbie-portia: As discussed, I've now removed value_schema

emmaportia (approved):

PR #357, opened 2025-04-25
Handle JSONSchema omissible fields in pydantic model generation

sam-portia:
# Description

Handles serialization of non-nullable omissible fields: fields that are not required in the JSON schema, but that are not nullable. Pydantic has no concept of an omitted field, so we must for it to be nullable and then make sure we don't serialize any None values.

See ticket desc for more info

Tested this with the Stripe example that was causing issues and no clarifications were raised, tools called correctly.

Ticket Link: https://linear.app/portialabs/issue/POR-1260/fix-for-optional-but-non-nullable-arguments-from-json-schema-incl-mcp 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/tool_registry.py
+++ b/portia/tool_registry.py
@@ -19,10 +19,11 @@ import asyncio
 import os
 import re
 from enum import StrEnum
-from typing import TYPE_CHECKING, Any, Callable, Literal, Union
+from typing import TYPE_CHECKING, Any, Callable, ClassVar, Literal, Union
 
 from jsonref import replace_refs
-from pydantic import BaseModel, Field, create_model
+from pydantic import BaseModel, Field, create_model, model_serializer
+from pydantic_core import PydanticUndefined
 
 from portia.cloud import PortiaCloudClient
 from portia.errors import DuplicateToolError, ToolNotFoundError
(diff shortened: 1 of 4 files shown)


portia/tool_registry.py
```diff
@@ -486,6 +488,31 @@ def __init__(self, config: Config) -> None:
         super().__init__(tools)
 
 
+class GeneratedBaseModel(BaseModel):
+    """BaseModel that is generated from a JSON schema.
+
+    Handles serialization of fields that must omit None values: fields that are not required in
+    the JSON schema, but that are not nullable. Pydantic has no concept of an omissible field,
+    so we must for it to be nullable and then make sure we don't serialize None values.
```
robbie-portia, line 496: nit: typo here - we must <WHAT> for it to be nullable

portia/tool_registry.py
```diff
@@ -506,30 +533,59 @@ def generate_pydantic_model_from_json_schema(
...
+        return type(None) in get_args(python_type)
+    return False
 
 
 def _generate_field(
     field_name: str,
     field: dict[str, Any],
     *,
     required: bool,
+    force_nullable: bool,
```
robbie-portia, line 578: I think it's worth a comment to explain what this is doing here

robbie-portia (approved):
Looking good!

PR #361, opened 2025-04-28
Fix latest release tagging

robbie-portia:
# Description

In terms of testing, would it be ok to create a new release tag and then delete it?

## Type of change

(select all that apply)

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/.github/workflows/create-release-tag.yml
+++ b/.github/workflows/create-release-tag.yml
@@ -5,10 +5,11 @@ on:
     types: [closed]
     branches:
       - main
+  workflow_dispatch:
 
 jobs:
   create-tag:
-    if: github.event.pull_request.merged == true && startsWith(github.event.pull_request.head.ref, 'release/v')
+    #if: github.event.pull_request.merged == true && startsWith(github.event.pull_request.head.ref, 'release/v')
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release-tag.yml
```diff
@@ -5,15 +5,17 @@ on:
     types: [closed]
     branches:
       - main
+  workflow_dispatch:
 
 jobs:
   create-tag:
-    if: github.event.pull_request.merged == true && startsWith(github.event.pull_request.head.ref, 'release/v')
+    #if: github.event.pull_request.merged == true && startsWith(github.event.pull_request.head.ref, 'release/v')
```
TomSPortia, line 12: Shall we delete rather than comment?

TomSPortia (approved):

TomSPortia:
In terms of testing, would it be ok to create a new release tag and then delete it? as long as it doesn't trigger a PyPi release I think thats fine?

.github/workflows/create-release-tag.yml
robbie-portia: Ah sorry, this is my bad - we want to keep this, I'd just pushed to try and test without pushing a new tag - now updated

PR #362, opened 2025-04-28
[POR-1317] Added introspection agent prompt improvements

OmarEl-Mohandes:
# Description

- Updated the prompt for introspection agent to work on specific data from plan and plan_run.
    - Extracted things likes the condition to evaluate, steps (in pretty print), outputs only ..etc.
    - Give more priority to outcomes (e.g. `COMPLETE` for some cases vs `SKIP`). 
    - I couldn't find meaningful cases to `FAIL` vs `COMPLETE` in conditionals. The only case we have now if is we're missing outputs from previous steps that are crucial to the step (and its condition), in that case it'll FAIL if that step is critical to the remaining steps. I think we will iterate more on this case later when there is more user feedback.
    - Based on the updated evals here: https://github.com/portiaAI/platform/pull/398 - Our score for o3-mini is up to 0.95 and claude is ~0.86. There might be some room for improvement, but as of now it's much more stable than before and more performant. 
- Updated the output of the variable in case it was skip/fail/complete, to be more human readable based on feedback.   

Ticket Link: N/A 

## Type of change

(select all that apply)

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [x] Refactor
- [ ] Documentation update

## Screenshots

## Changelog


tests/unit/test_portia.py
```diff
@@ -1166,14 +1169,14 @@ def test_handle_introspection_outcome_fail(portia: Portia) -> None:
         state=PlanRunState.IN_PROGRESS,
     )
 
-    # Mock the introspection agent to return FAIL
+
```
robbie-portia, line 1172: mega nit: We can probably remove these double new lines

portia/introspection_agents/default_introspection_agent.py
```diff
@@ -45,33 +45,57 @@ def __init__(self, config: Config, agent_memory: AgentMemory) -> None:
...
+step B ONLY if Step A output is included in the inputs list for Step B or
+mentioned in the condition attribute for step B.
+- Steps CAN NOT get executed on partial information. A step will only be executed if
+all inputs are present and its condition is true (if the step has condition).
+- Your outcome will not affect any future plan runs.
+
+Provide an outcome from the following list (ordered by preference):
+  1- COMPLETE -> stops the execution and does not execute remaining steps.
+   - You should favour this outcome if no steps remaining in the list
+    that  can be executed without the current step output.
```
robbie-portia, line 67:
```suggestion
    that can be executed without the current step output.
```

portia/introspection_agents/default_introspection_agent.py
```diff
@@ -45,33 +45,57 @@ def __init__(self, config: Config, agent_memory: AgentMemory) -> None:
...
+   - You should favour this outcome if the condition of this step applies
+   to all remaining steps.
+  2- SKIP -> ONLY skips the current step and continue executing next steps.
+  3- FAIL -> fail and stops the execution of all the remaining steps in the plan.
```
robbie-portia, line 71: As discussed, let's remove this.

portia/introspection_agents/default_introspection_agent.py
```diff
@@ -45,33 +45,57 @@ def __init__(self, config: Config, agent_memory: AgentMemory) -> None:
...
+  4- CONTINUE -> continue execution for the current step.
```
robbie-portia, line 72: I wonder if we should say something about defaulting to this if you aren't sure?

portia/introspection_agents/default_introspection_agent.py
```diff
@@ -45,33 +45,57 @@ def __init__(self, config: Config, agent_memory: AgentMemory) -> None:
...
+"""
                     ),
                 ),
                 HumanMessagePromptTemplate.from_template(
                     "Today's date is {current_date} and today is {current_day_of_week}.\n"
                     "Review the following plan + current PlanRun.\n"
-                    "Current Plan: {plan}\n"
-                    "Current PlanRun: {plan_run}\n"
+                    "The condition to evaluate is: {condition}.\n"
+                    "The current step index is: {current_step_idex}.\n"
```
robbie-portia, line 94: You could say "we are at step {} out of {} in total"

robbie-portia (commented):
Looks good

portia/introspection_agents/default_introspection_agent.py
OmarEl-Mohandes: Hmmm.. I'm a bit worry about doing this, as it might nudge to reason towards being not sure and short cut to this, rather than think more. In the guides below, we say otherwise continue (hopefully that would be enough).

portia/introspection_agents/default_introspection_agent.py
```diff
@@ -45,33 +45,52 @@ def __init__(self, config: Config, agent_memory: AgentMemory) -> None:
...
+- Your outcome will not affect any future plan runs.
+
+Provide an outcome from the following list (ordered by preference):
+  1- COMPLETE -> stops the execution and does not execute remaining steps.
+   - You should favour this outcome if no steps remaining in the list
+    that can be executed without the current step output.
+   - You should favour this outcome if the condition of this step applies
+   to all remaining steps.
+  2- SKIP -> ONLY skips the current step and continue executing next steps.
+  4- CONTINUE -> continue execution for the current step.
```
robbie-portia, line 71:
```suggestion
  3- CONTINUE -> continue execution for the current step.
```

robbie-portia (approved):
LGTM

PR #371, opened 2025-04-30
Raise ready status clarifications up-front and after any tool-run clarifications

sam-portia:
# Description

The goal is to check for any clarifications we can possibly detect and raise all of them as early as possible to avoid repeated interruptions to the flow.

Currently, the only type of clarification that can be detected ahead of time is tool-readiness (i.e. Auth).

More specifically it will now:
* Check all remaining steps for tool ready status and raise all resulting clarifications up-front (start of `resume`)
* Check all remaining steps for tool ready status after a clarification is raised from a task for any other reason

Ticket Link: https://linear.app/portialabs/issue/POR-1231/pull-the-portia-cloud-tool-auth-forward-to-before-planruns-execute

> [!WARNING]
> Need to think carefully about the release of this change as it currently depends on the platform being released first (with the update to the ready endpoint - [PR](https://github.com/portiaAI/platform/pull/410))
> GIven the platform prod release currently is dependent on already-merged SDK changes, it doesnt seem a good idea to create a bi-directional dependency.
> For that reason, the simplest thing would be to wait until after the next prod release for this PR to get merged.

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [x] Requires sync with platform release
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/portia.py
+++ b/portia/portia.py
@@ -1027,7 +1027,6 @@ class Portia:
                     existing_clarification_ids = {c.id for c in plan_run.outputs.clarifications}  # pyright: ignore[reportUnhashable]
                     for clarification in ready_response.clarifications:
                         if clarification.id not in existing_clarification_ids:
-                            clarification.step = None  # Update backend to return None for ready
                             plan_run.outputs.clarifications.append(clarification)
                             new_clarifications_raised = True
                             logger().info(
--- a/portia/portia.py
+++ b/portia/portia.py
@@ -336,11 +336,9 @@ class Portia:
         plan = self.storage.get_plan(plan_id=plan_run.plan_id)
 
         # Perform initial readiness check
-        self._check_remaining_tool_readiness(plan, plan_run)
-
-        # If the readiness check resulted in needing clarification, return immediately
-        if plan_run.state == PlanRunState.NEED_CLARIFICATION:
-            logger().debug("Returning plan run early due to readiness check clarifications.")
+        if len(ready_clarifications := self._check_remaining_tool_readiness(plan, plan_run)):
+            self._raise_clarifications(ready_clarifications, plan_run)
+            self._handle_clarifications(plan_run)
             return plan_run
 
         # if the run has execution context associated, but none is set then use it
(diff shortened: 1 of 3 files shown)


portia/portia.py
```diff
@@ -1364,6 +1422,64 @@ def _persist_step_state(self, plan_run: PlanRun, step: Step) -> None:
...
+                ready_response = tool.ready(tool_run_context)
+                if not ready_response.ready:
+                    ready_clarifications.extend(ready_response.clarifications)
+
+        if len(portia_cloud_tool_ids_remaining) == 0:
+            return ready_clarifications
+
+        portia_tools_ready_response = PortiaRemoteTool.batch_ready_check(
+            self.config,
+            list(portia_cloud_tool_ids_remaining),
```
OmarEl-Mohandes, line 1475: You can just pass here the set instead of sorting it later. Also, to enforce sending unique set of ids from the SDK.

portia/portia.py
```diff
@@ -1036,9 +1060,29 @@ def _execute_plan_run(self, plan: Plan, plan_run: PlanRun) -> PlanRun:  # noqa:
...
+                )
+                > 0
+            ):
+                # If execution raised a clarification, re-check readiness of subsequent tools
+                logger().debug(
+                    "Re-checking tool readiness after execution clarification.",
+                    plan=str(plan.id),
+                    plan_run=str(plan_run.id),
+                )
+                ready_clarifications = self._check_remaining_tool_readiness(
```
OmarEl-Mohandes, line 1078: It's nice that re-check again. I'm wondering if it's worth checking the type of clarifications to avoid sending requests to the backend. Don't feel strong about this atm, happy to wait and see if it's worth it.

portia/portia.py
```diff
@@ -1036,9 +1060,29 @@ def _execute_plan_run(self, plan: Plan, plan_run: PlanRun) -> PlanRun:  # noqa:
...
+                    plan,
+                    plan_run,
+                    start_index=index + 1,
+                )
+                combined_clarifications = new_clarifications + ready_clarifications
```
OmarEl-Mohandes, line 1083: Should we start the index from current instead of combining old clarification? .. Consider this case:

 - google calendar tool -> current_step (raised clarification)
 - gmail (will raise a clarification for this and all remain google tools)

Do you think we'll end up with multiple clarifications in this case for the same client?

sam-portia: Thats a good point - If the clarification raised is an auth one, we should combine that into a batch call..

Whats a bit nasty is that we can't actually be sure that a clarification was raised because of auth? We can check for an Action clarification, but that doesnt guarantee its an Auth clarification... I suppose for PortiaRemoteTools this isnt so much of a problem

portia/portia.py
sam-portia: true - will update

portia/portia.py
sam-portia: Done this now

portia/portia.py
```diff
@@ -1036,9 +1060,42 @@ def _execute_plan_run(self, plan: Plan, plan_run: PlanRun) -> PlanRun:  # noqa:
...
+                    )
+                    and new_clarifications[0].category == ClarificationCategory.ACTION
+                ):
+                    combined_clarifications = self._check_remaining_tool_readiness(
+                        plan,
+                        plan_run,
+                        start_index=index,
+                    )
+                else:
+                    ready_clarifications = self._check_remaining_tool_readiness(
```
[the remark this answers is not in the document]
sam-portia: Yeah I plan to follow up after the first release with playing around with my own custom tool readiness to see how it feels, I'm sure it could be improved. It should be OK as long as users ensure their ready() method can be called multiple times though

PR #384, opened 2025-05-05
Allow running a plan by ID

robbie-portia:
# Description

Allow running a plan from ID rather than from the full plan

## Type of change

(select all that apply)

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/portia.py
+++ b/portia/portia.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 import time
 from importlib.metadata import version
 from typing import TYPE_CHECKING
+from uuid import UUID
 
 from portia.clarification import (
     Clarification,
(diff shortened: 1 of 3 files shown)


portia/portia.py
```diff
@@ -277,14 +278,14 @@ def plan(
 
     def run_plan(
         self,
-        plan: Plan,
+        plan: Plan | PlanUUID,
```
Nathanjp91, line 281: | UUID and convert to planuuid here as an option as well. Not many peole are going to use planUUIDs under the hood. and if they're retrieving from their own storage, they might be using just the UUID.

robbie-portia: Good call - now updated

Nathanjp91 (approved):
LGTM

PR #407, opened 2025-05-12
Updates from plan input dogfooding

robbie-portia:
# Description

Changes after our plan inputs dogfooding session:
* Allow calling public methods with dicts
* Change PlanInput to have an optional value and use lists of them
* A few spelling tweaks in prompts
* Change plan field from `inputs` to `plan_inputs`
* Add an indent to plan pretty printing

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/portia.py
+++ b/portia/portia.py
@@ -31,6 +31,7 @@ from portia.clarification import (
     ClarificationCategory,
 )
 from portia.cloud import PortiaCloudClient
+from portia.common import Serializable
 from portia.config import (
     Config,
     ExecutionAgentType,
@@ -82,7 +83,6 @@ from portia.tool_registry import (
 from portia.tool_wrapper import ToolCallWrapper
 
 if TYPE_CHECKING:
-    from portia.common import Serializable
     from portia.execution_agents.base_execution_agent import BaseExecutionAgent
     from portia.planning_agents.base_planning_agent import BasePlanningAgent
 
@@ -161,7 +161,7 @@ class Portia:
         tools: list[Tool] | list[str] | None = None,
         example_plans: list[Plan] | None = None,
         end_user: str | EndUser | None = None,
-        plan_run_inputs: dict[PlanInput, Serializable] | None = None,
+        plan_run_inputs: list[PlanInput] | list[dict[str, str]] | dict[str, str] | None = None,
     ) -> PlanRun:
         """End-to-end function to generate a plan and then execute it.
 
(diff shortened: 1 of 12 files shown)


tests/unit/test_portia.py
```diff
@@ -1291,19 +1297,40 @@ def test_portia_initialize_end_user(portia: Portia) -> None:
     assert storage_end_user.name == "Bob Smith"
 
 
-def test_portia_run_with_plan_run_inputs(portia: Portia, planning_model: MagicMock) -> None:
-    """Test that Portia.run handles plan inputs correctly."""
-    num_a_input = PlanInput(name="$num_a", description="Number A")
-    num_b_input = PlanInput(name="$num_b", description="Number B")
-
+@pytest.mark.parametrize(
```
Nathanjp91, line 1300: Good use of parametrize test here.

portia/portia.py
```diff
@@ -263,25 +302,55 @@ def plan(
...
+                            name=plan_input["name"],
+                            description=plan_input.get("description", None),
+                        )
+                    )
+                elif isinstance(plan_input, str):
+                    to_return.append(PlanInput(name=plan_input))
+                else:
+                    to_return.append(plan_input)
+            return to_return
+        return None
```
Nathanjp91, line 328: I wonder if we should raise here instead, because we've already captured if plan_inputs is None, so if they reach this point it means they've passed something along to this function and returning None here might return nothing silently and cause potential issues upstream, where a Raise TypeError("Expected list[PlanInput], ...") would fail faster and highlight issues.

portia/portia.py
```diff
@@ -174,26 +174,61 @@ def run(
...
+                    )
+                else:
+                    to_return.append(plan_run_input)
+            return to_return
+        if isinstance(plan_run_inputs, dict):
+            to_return = []
+            for key, value in plan_run_inputs.items():
+                to_return.append(PlanInput(name=key, value=value))
+            return to_return
+        return None
```
Nathanjp91, line 223: As the other, raise here instead imo.

Nathanjp91 (changes requested):
Mostly good, Minor change with coercion raising I think is cleaner.

portia/portia.py
robbie-portia: Good point - now updated!

Nathanjp91 (approved):
LGTM

PR #408, opened 2025-05-12
Add plan verification + retries to planner

robbie-portia:
# Description

* Add a check to the planner that checks the inputs to each step are valid
* Add retries to the planner on tool or input errors

## Type of change

(select all that apply)

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog


Nathanjp91 (approved):
Looks good. We might want to think about capturing this information in the form of an Eval though. Like yes, it'll likely result in less issues with planner, but also how many retries are we using is useful information. Can we capture that in a linear ticket if you agree?

robbie-portia:
@Nathanjp91 - yeah, good call. I'll have a look at adding it now and if it takes more than 10 mins I'll raise a ticket for it.

PR #413, opened 2025-05-12
[RUN-44] Update llm_tool to handle variaty of inputs to make use of memory agent

OmarEl-Mohandes:
# Description

The current input for the `llm_tool` is forcing gemini model (2.0 flash) to try to parse the big input (search results) into list of str in tool agent node, which is causing `MALFORMED_FUNCTION_CALL` ([example](https://smith.langchain.com/o/e2eb5117-7ff4-446a-b9ac-b768f33886e1/projects/p/500223c8-b833-4853-9ef6-0016c236461a?timeModel=%7B%22duration%22%3A%227d%22%7D&peek=236f0c1b-0a8f-4471-851a-25fb629ba829&peeked_trace=ff9fa307-5d94-475e-a15a-01ee89d90c59)) for a lot of cases because of tavily search results. We need to fix this for 2.0 flash because they can't upgrade to gemini 2.5 atm.

### Fix 
I relaxed the input types for `llm_tool` to accept it as `List[Any] | str | None` instead of `List[str]` only, then handle casting it back internally to list of str in the tool itself. 

This has forced the gemini model (2.0 flash) to use input_variable name from agent memory and avoid parsing the input itself when producing tool calling for llm_tool ([example](https://smith.langchain.com/o/e2eb5117-7ff4-446a-b9ac-b768f33886e1/projects/p/500223c8-b833-4853-9ef6-0016c236461a?timeModel=%7B%22duration%22%3A%227d%22%7D&peek=f4d265cc-f0d5-4cfb-878b-2497fe8afc78&peeked_trace=c6877734-7d2b-49b8-a1e2-f9eaf10990a5)). 

1) The response is much faster for the tool calling agent node (thanks to agent memory) and ;
2) Reducing the chance of getting `MALFORMED_FUNCTION_CALL` (got 100% accuracy for couple for runs consecutively). Also tested with openai models and got the same results.
 

Ticket Link:  https://linear.app/portialabs/issue/RUN-44/tunic-issues-with-running-long-plan
## Type of change

(select all that apply)

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/open_source_tools/llm_tool.py
+++ b/portia/open_source_tools/llm_tool.py
@@ -95,7 +95,11 @@ class LLMTool(Tool[str]):
             "tasks"
         )
 
-        task_data_str = "\n".join(task_data_list)
+        task_data_str = (
+            "\n".join(task_data_list)
+            if isinstance(task_data_list, list)
+            else str(task_data_list)
+        )
 
         task_str = task
         if task_data_str:
--- a/portia/open_source_tools/llm_tool.py
+++ b/portia/open_source_tools/llm_tool.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 from typing import Any, ClassVar
 
-from pydantic import BaseModel, Field, field_validator
+from pydantic import BaseModel, Field
 
 from portia.model import GenerativeModel, Message
 from portia.tool import Tool, ToolRunContext
@@ -25,19 +25,6 @@ class LLMToolSchema(BaseModel):
         "Important: This should include all relevant data in their entirety, "
         "from the first to the last character (i.e. NOT a summary).",
     )
-
-    @field_validator("task_data", mode="before")
-    def validate_task_data(cls, v) -> list[str]:  # noqa: ANN001, N805
-        """Convert task_data to appropriate format."""
-        if v is None:
-            return []
-
-        if isinstance(v, str):
-            return [v]
-
-        if isinstance(v, list):
-            return [str(item) for item in v]
-        return [str(v)]
 
 
 class LLMTool(Tool[str]):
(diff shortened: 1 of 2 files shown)


portia/open_source_tools/llm_tool.py
```diff
@@ -69,8 +81,11 @@ class LLMTool(Tool[str]):
         "the model will be resolved from the config.",
     )
 
-    def run(self, ctx: ToolRunContext, task: str, input_data: Sequence[str] = ()) -> str:
+    def run(self, ctx: ToolRunContext, task: str, task_data: list[Any] | str | None = None) -> str:
         """Run the LLMTool."""
+        # Use the schema validator to handle the task_data conversion
+        task_data_list = LLMToolSchema(task=task, task_data=task_data).task_data
```
robbie-portia, line 87: Rather than having the schema validator on the model, and then still having to deal with the fact it might not be a list further down when creating task_data_str, would it be easier to just have a method that took in task_data, did the type validation, then did the string joining and returned task_data_str?

portia/open_source_tools/llm_tool.py
```diff
@@ -20,13 +17,28 @@ class LLMToolSchema(BaseModel):
         ...,
         description="The task to be completed by the LLM tool",
     )
-    input_data: list[str] = Field(
-        default_factory=list,
-        description="A list of strings containing relevant data that should be used to complete "
-        "the task. Important: This should include all relevant data in their entirety, from the "
-        "first to the last character (i.e. NOT a summary).",
+    task_data: list[Any] | str | None = Field(
```
robbie-portia, line 20: Do you know if there's any reason why we originally went for list[str]?

OmarEl-Mohandes: Good question. This was in response for the incident when LLM tool ignored inputs. Luckly, Sam mentioned here [the context](https://docs.google.com/document/d/1p3I8M9YnL5w2o1JMmFevbGrCHqWH0b0PiLY_bpSuTJE/edit?tab=t.0). The goal was for LLM tool to be able to accept more input values rather than stuffing everything in task.

robbie-portia (approved):
Looking good!

PR #415, opened 2025-05-13
Plan input bug fix

robbie-portia:
# Description

The name changed on the backend but wasn't changed here, so we weren't receiving plan inputs from the backend. I've also updated the test so it would have caught this.

## Type of change

(select all that apply)

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/tests/integration/test_e2e.py
+++ b/tests/integration/test_e2e.py
@@ -594,7 +594,6 @@ def test_plan_inputs() -> None:
 
     config = Config.from_default(
         default_log_level=LogLevel.DEBUG,
-        storage_class=StorageClass.MEMORY,
     )
     portia = Portia(config=config, tools=ToolRegistry([AdditionTool()]))
     plan = portia.plan(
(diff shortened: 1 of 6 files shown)


Nathanjp91 (approved):

tests/integration/test_e2e.py
```diff
@@ -594,7 +594,6 @@ class AdditionNumbers(BaseModel):
 
     config = Config.from_default(
         default_log_level=LogLevel.DEBUG,
-        storage_class=StorageClass.MEMORY,
```
emmaportia, line 597: why be changing this?

emmaportia (approved):

tests/integration/test_e2e.py
robbie-portia: So we use cloud storage in this integration test which would have caught the issue

PR #439, opened 2025-05-19
Sort execution agent email fussiness

robbie-portia:
# Description

Interestingly, adding explanation before it chooses the made_up field essentially works as a chain of thought, which helps it get the right answer. Adding the tool description also helps.
I've also included a couple of end user fixes - these are what is needed to get https://github.com/portiaAI/platform/pull/462 in (occasionally things break because the agent thinks there is an issue when it is passed an empty end user bock).

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog


OmarEl-Mohandes (approved):
Interesting one on the explanation! Would be nice if you can update the lock for evals and steel thread for these changes.

PR #442, opened 2025-05-20
Async tool

TomSPortia:
# Description

First part of the changes for async in the SDK - allowing tools to be async

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/open_source_tools/browser_tool.py
+++ b/portia/open_source_tools/browser_tool.py
@@ -350,7 +350,7 @@ class BrowserToolForUrl(BrowserTool):
             infrastructure_option=infrastructure_option,
         )
 
-    async def run_async(
+    async def run_async(  # type: ignore  # noqa: PGH003
         self, ctx: ToolRunContext, task: str
     ) -> str | BaseModel | ActionClarification:
         """Run the BrowserToolForUrl."""
--- a/portia/open_source_tools/browser_tool.py
+++ b/portia/open_source_tools/browser_tool.py
@@ -257,6 +257,7 @@ class BrowserTool(Tool):
         )
 
         result = await agent.run()
+        print(result)
         task_result = output_model.model_validate(json.loads(result.final_result()))  # type: ignore reportCallIssue
 
         if task_result.human_login_required:  # type: ignore reportCallIssue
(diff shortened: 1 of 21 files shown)


portia/tool.py
```diff
@@ -130,6 +128,14 @@ class Tool(BaseModel, Generic[SERIALIZABLE_TYPE_VAR]):
         "Tools may not require a summary if they already produce a nice textual output.",
     )
 
+    def __init_subclass__(cls) -> None:
```
TomSPortia, line 131: General approach is to allow tools to now either implement `def run` or `async def run_async` so its backwards compatible but you only have to provide one implementation,

Nathanjp91 (commented):
Given that we're swapping out abstractmethod for an init_subclass approach, can you add in a test to make sure that they can't be created without defining a run or an async run function?

portia/open_source_tools/browser_tool.py
```diff
@@ -352,9 +345,9 @@ def __init__(  # noqa: PLR0913
             infrastructure_option=infrastructure_option,
         )
 
-    def run(self, ctx: ToolRunContext, task: str) -> str | ActionClarification:  # type: ignore reportIncompatibleMethodOverride
+    async def run_async(self, ctx: ToolRunContext, task: str) -> str | ActionClarification:  # type: ignore reportIncompatibleMethodOverride
         """Run the BrowserToolForUrl."""
-        return super().run(ctx, self.url, task)  # pragma: no cover
+        return await super().run_async(ctx, self.url, task)  # pragma: no cover
```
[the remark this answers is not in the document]
TomSPortia: nope, removed

portia/tool.py
```diff
@@ -197,8 +234,25 @@ def _run(
 
         """
         try:
-            output = self.run(ctx, *args, **kwargs)
+
+            def is_overridden(method_name: str) -> bool:
```
Nathanjp91, line 238: Thoughts on making these properties on Tool?

```python
if tool.is_async:
    tool.run_async(...)
else: # we know it has to have at least run_async or run
    tool.run(...)
```
The code here feels a bit hard to reason about, and a property on the tool to do this check might be useful in other ways while making it clear that's what you're checking.

Nathanjp91 (commented):
Thoughts on property for async check? Happy to discuss

PR #443, opened 2025-05-20
[POR-1320] Upgrade execution models and fix planning/introspection for anthropic

OmarEl-Mohandes:
# Description

Based on the investigation (see doc in drive) we should upgrade our default models for execution agent for openai and anthropic. Also fix planning and introspection to have same version as our evals, since it's clearly performing better.

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/.github/workflows/evals.yml
+++ b/.github/workflows/evals.yml
@@ -104,7 +104,7 @@ jobs:
           PORTIA_API_KEY: ${{ secrets.PORTIA_EVAL_API_KEY }}
           PORTIA_API_ENDPOINT: ${{ secrets.PORTIA_EVAL_API_ENDPOINT }}
         run: |
-          EVAL_OUTPUT=$(uv run cli.py agent eval --slice_name=main --model=anthropic/claude-3-5-sonnet-latest --threshold_file=execution/thresholds/claude-3-5-sonnet-latest/thresholds_local.yaml --reps 1 --metadata "pr=${{ github.event.pull_request.number }},author=${{ github.event.pull_request.user.login || github.actor }},run=${{ github.event_name == 'push' && 'main' || 'pr' }},env=local,repo=sdk"  --max_concurrency 32)
+          EVAL_OUTPUT=$(uv run cli.py agent eval --slice_name=main --model=anthropic/claude-3-7-sonnet-latest --threshold_file=execution/thresholds/claude-3-5-sonnet-latest/thresholds_local.yaml --reps 1 --metadata "pr=${{ github.event.pull_request.number }},author=${{ github.event.pull_request.user.login || github.actor }},run=${{ github.event_name == 'push' && 'main' || 'pr' }},env=local,repo=sdk"  --max_concurrency 32)
           echo $EVAL_OUTPUT
           echo "eval_url=$(echo "$EVAL_OUTPUT" | grep -o 'https://smith.langchain.com/.*')" >> $GITHUB_OUTPUT
           echo "eval_name=$(echo "$EVAL_OUTPUT" | grep -oP "experiment:\s*'\K[^']+")" >> $GITHUB_OUTPUT
--- a/portia/config.py
+++ b/portia/config.py
@@ -122,6 +122,7 @@ class LLMModel(Enum):
 
     # OpenAI
     GPT_4_O = Model(provider=LLMProvider.OPENAI, model_name="gpt-4o")
+    GPT_4_1 = Model(provider=LLMProvider.OPENAI, model_name="gpt-4.1")
     GPT_4_O_MINI = Model(provider=LLMProvider.OPENAI, model_name="gpt-4o-mini")
     GPT_3_5_TURBO = Model(provider=LLMProvider.OPENAI, model_name="gpt-3.5-turbo")
     O_3_MINI = Model(provider=LLMProvider.OPENAI, model_name="o3-mini")
@@ -152,6 +153,7 @@ class LLMModel(Enum):
     # Azure OpenAI
     AZURE_GPT_4_O = Model(provider=LLMProvider.AZURE_OPENAI, model_name="gpt-4o")
     AZURE_GPT_4_O_MINI = Model(provider=LLMProvider.AZURE_OPENAI, model_name="gpt-4o-mini")
+    AZURE_GPT_4_1 = Model(provider=LLMProvider.AZURE_OPENAI, model_name="gpt-4.1")
     AZURE_O_3_MINI = Model(provider=LLMProvider.AZURE_OPENAI, model_name="o3-mini")
 
     @property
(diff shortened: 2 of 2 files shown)


.github/workflows/evals.yml
```diff
@@ -104,7 +104,7 @@ jobs:
           PORTIA_API_KEY: ${{ secrets.PORTIA_EVAL_API_KEY }}
           PORTIA_API_ENDPOINT: ${{ secrets.PORTIA_EVAL_API_ENDPOINT }}
         run: |
-          EVAL_OUTPUT=$(uv run cli.py agent eval --slice_name=main --model=anthropic/claude-3-5-sonnet-latest --threshold_file=execution/thresholds/claude-3-5-sonnet-latest/thresholds_local.yaml --reps 1 --metadata "pr=${{ github.event.pull_request.number }},author=${{ github.event.pull_request.user.login || github.actor }},run=${{ github.event_name == 'push' && 'main' || 'pr' }},env=local,repo=sdk"  --max_concurrency 32)
+          EVAL_OUTPUT=$(uv run cli.py agent eval --slice_name=main --model=anthropic/claude-3-7-sonnet-latest --threshold_file=execution/thresholds/claude-3-5-sonnet-latest/thresholds_local.yaml --reps 1 --metadata "pr=${{ github.event.pull_request.number }},author=${{ github.event.pull_request.user.login || github.actor }},run=${{ github.event_name == 'push' && 'main' || 'pr' }},env=local,repo=sdk"  --max_concurrency 32)
```
OmarEl-Mohandes, line 107: I'll come back to change the threshold folder to be `claude-3-5-sonnet-latest`, but will raise a PR later once backend changes are fully in prod.

portia/config.py
```diff
@@ -767,27 +769,35 @@ def _parse_model_string(self, model_string: str) -> GenerativeModel:
         - google (requires portia-sdk-python[google] to be installed)
         - azure-openai
 
+        The optional "reasoning" parameter enables reasoning mode for models that support it.
+
         Args:
-            model_string (str): The model string to parse. E.G. "openai/gpt-4o"
+            model_string (str): The model string to parse. E.G. "openai/gpt-4o" or "anthropic/claude-3-7-sonnet-latest/reasoning"
```
sam-portia, line 775: Im not 100% convinced about encoding the reasoning setting into the model string in this way.. I can see that its attractive, and its kind of neat but I think it has some drawbacks:
- It adds a Portia-proprietary syntax that might confuse people. The model name being 1-1 with the names used in the LLM Provider's API makes it easy to get use to our code.
- You can add `/reasoning` to any model name and we'll ingest it (e.g. `openai/gpt-4o/reasoning`) despite many models not having a reasoning capability.
- What if there was a model name that legitimately had a `/` in it? Our code would start to get very confused


We do have a problem that you can't pass in model specific / provider specific configuration settings, I think I would solve for that instead?

e.g.
```python
Config(
    model_settings={
        "anthropic/claude-3-7-sonnet-latest": {
            "thinking": {"type": "enabled", "budget_tokens": 2000}
        },
        "openai/o3-mini": {"reasoning_effort": "medium"},
    }
)
```
Where whatever is in this dict is injected into the GenerativeModel class kwargs when instantiating the object for that model

Something along those lines should solve the problem and also be more generic and useful for other model parameters too

.github/workflows/evals.yml
```diff
@@ -36,7 +36,10 @@ jobs:
         with:
           path: portia-sdk-python
           repository: portiaAI/portia-sdk-python
-          ref: ${{ github.head_ref || 'main' }}
+          # For push events, use github.ref_name (branch name that was pushed)
```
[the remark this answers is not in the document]
OmarEl-Mohandes: Yes, we have enable push on main only for this workflow.

portia/config.py
sam-portia: Its something that I thought about as part of LLM agnosticism & configuration changes but it was descoped, so it would be really nice!

As an aside (and you probably don't need it here so may not be worth doing) there could also be a set of provider specific settings (e.g. applying to all models under that provider). Something along the lines of:

```python
Config(provider_settings={LLMProvider.ANTHROPIC: {"temperature": 0.5}})
```

sam-portia: Right yeah, having the config for models be per agent makes things more complicated

I think then that the best approach might be to pass explicit GenerativeModel instances?

e.g. 
```python
Config.from_default(
    planning_model=AnthropicGenerativeModel(
        model_name="claude-3-7-sonnet-latest",
        thinking={"type": "enabled", "budget_tokens": 3000},
    )
)
```

sam-portia: Also Claude 4 was just announced 😁

OmarEl-Mohandes: Mmm..let me try this out, seems easier to grasp. 

On claude 4:

portia/model.py
```diff
@@ -141,16 +140,26 @@ def get_structured_response(
...
+            return False
+        return (
+            self.__class__ == other.__class__
+            and self.provider == other.provider
+            and self.model_name == other.model_name
+        )
+
+    def __hash__(self) -> int:
+        """Generate hash based on class, provider, and model name."""
+        return hash((self.__class__, self.provider, self.model_name))
```
sam-portia, line 161: I wouldn't implement hash and eq for these types given there are other model parameters that mean 2 instances of the class with the same provider and model name should be considered different.

If you need these methods for testing - you could convert to string before comparison? That encodes the model name and provider (and provider is 1-1 with class)

portia/config.py
```diff
@@ -762,14 +804,19 @@ def _parse_model_string(self, model_string: str) -> GenerativeModel:
...
```
sam-portia, line 810: This docstring is out of date now?

portia/config.py
```diff
@@ -762,14 +804,19 @@ def _parse_model_string(self, model_string: str) -> GenerativeModel:
...
 
         Returns:
             GenerativeModel: The parsed model.
 
-        """
-        provider, model_name = model_string.strip().split("/", maxsplit=1)
+        """  # noqa: E501
+        parts = model_string.strip().split("/", maxsplit=2)
```
sam-portia, line 816: And this part - we can revert?

portia/config.py
```diff
@@ -753,7 +795,7 @@ def get_generative_model(
         return model
 
     def _parse_model_string(self, model_string: str) -> GenerativeModel:
-        """Parse a model string in the form of "provider-prefix/model_name` to a GenerativeModel.
+        """Parse a model string in the form of "provider-prefix/model_name[/reasoning]` to a GenerativeModel.
```
sam-portia, line 798: Also out of date?

sam-portia (approved):
Some minor cleanups potentially, but otherwise LGTM!

PR #475, opened 2025-05-29
Browser tool - enable multiple steps within the same plan run

robbie-portia:
# Description

Enables multiple steps using the browser tool with the same session in the same plan run.

## Type of change

(select all that apply)

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog


portia/open_source_tools/browser_tool.py
```diff
@@ -207,11 +207,13 @@ class BrowserTool(Tool[str | BaseModel]):
...
         init_var=True,
         default=(
-            "General purpose browser tool. Can be used to navigate to a URL and "
-            "complete tasks. Should only be used if the task requires a browser "
-            "and you are sure of the URL. This tool handles a full end to end task. "
-            "DO NOT break a single task into multiple browser tool steps as each step is run "
-            "in an independent new session!"
+            "General purpose browser tool. Can be used to navigate to a URL and complete tasks. "
+            "Should only be used if the task requires a browser and you are sure of the URL. "
+            "This tool handles a full end to end task. It is capable of doing multiple things "
```
robbie-portia, line 212: What do you think of this encouragement to do things in a single step? I'm on the fence about it. When I was playing around with the tool, without this in the prompt a query like "Go to the users's linkedin messages and send a message saying hi to Emma Burrows" would result in 2 browser-tool steps rather than one. That's not really an issue - it has the benefit of ensuring we stick closely to the plan, but has the drawback of needing to go through the execution agent again.

portia/open_source_tools/registry.py
```diff
@@ -1,6 +1,5 @@
 """Example registry containing simple tools."""
 
-import logging
```
robbie-portia, line 3: This was just removing an unused logger (and making sure all our loggers use the portia logger)

portia/open_source_tools/browser_tool.py
emmaportia: agree, i think i wouldn't explicitly encourage it to do single steps

portia/open_source_tools/browser_tool.py
```diff
@@ -742,3 +755,22 @@ def setup_browser(self, ctx: ToolRunContext) -> Browser:
                     cdp_url=session_connect_url,
                 ),
             )
+
+        def _is_final_browser_tool_call(self, plan_run: PlanRun, plan: Plan) -> bool:
```
emmaportia, line 759: nit: would inline this

emmaportia (approved):

robbie-portia:
I did a bit of testing to check it wasn't opening a new tab and I didn't see it do that, but it's something we should keep an eye on

Issue #466, opened 2025-05-28
[£60] Add the ability to wrap a method with an @tool decorator


emmaportia:
We'd like to be able to declare simpler tools more easily using a @tool decorator. Consider the below example for the weather tool:

```
class WeatherToolSchema(BaseModel):
    """Input for WeatherTool."""

    city: str = Field(..., description="The city to get the weather for")


class WeatherTool(Tool[str]):
    """Get the weather for a given city."""

    id: str = "weather_tool"
    name: str = "Weather Tool"
    description: str = "Get the weather for a given city"
    args_schema: type[BaseModel] = WeatherToolSchema
    output_schema: tuple[str, str] = ("str", "String output of the weather with temp and city")

    def run(self, _: ToolRunContext, city: str) -> str:
        """Run the WeatherTool."""
        api_key = os.getenv("OPENWEATHERMAP_API_KEY")
        if not api_key or api_key == "":
            raise ToolHardError("OPENWEATHERMAP_API_KEY is required")
        url = (
            f"http://api.openweathermap.org/data/2.5/weather?q={city}&appid={api_key}&units=metric"
        )
        response = httpx.get(url)
        response.raise_for_status()
        data = response.json()
        if "weather" not in data:
            raise ToolSoftError(f"No data found for: {city}")
        weather = data["weather"][0]["description"]
        if "main" not in data:
            raise ToolSoftError(f"No main data found for city: {city}")
        temp = data["main"]["temp"]
        return f"The current weather in {city} is {weather} with a temperature of {temp}°C."
```

Instead, this should be declared as:
```
@tool
def weather_tool(city: str) -> str:
    """Retrieves the weather of the provided city""""
    # Same implementation as the run method above
```

The implementation should:
- Use the documentation to populate the description
- Have an id and name based on the method name
- Populate the output schema based on the return type of the function
- Populate the input schema based on the inputs to the function (and arg documentation if available)

PR #519, opened 2025-06-13
Add feature tool decorator

jingkaihe:
# Description

add tool decorator for creating tools from functions

now you can create tool via

```python
# basic
@tool
def multiply_numbers(a: float, b: float) -> float:
    """Multiply two numbers together."""
    return a * b

# with python annotations + pydantic
@tool
def calculate_area(
    length: Annotated[float, Field(description="The length of the rectangle")],
    width: Annotated[float, Field(description="The width of the rectangle", gt=0)]
) -> float:
    """Calculate the area of a rectangle."""
    return length * width
```

Ticket Link: https://github.com/portiaAI/portia-sdk-python/issues/466

## Type of change

(select all that apply)

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

No screenshots taken but here is an example run and its output - https://gist.github.com/jingkaihe/99cf7db3f5fa8e5e40f9e1bead30dab2

## Changelog

### DIFF
--- a/portia/tool_decorator.py
+++ b/portia/tool_decorator.py
@@ -206,7 +206,7 @@ def _process_annotated_type(
         # Malformed Annotated, treat as Any
         param_type = Any
         description = _form_param_description(param_name, func_name)
-        field_info = cast(FieldInfo, Field(default=default, description=description))
+        field_info = cast("FieldInfo", Field(default=default, description=description))
         return param_type, field_info
 
     # First arg is the actual type
@@ -218,9 +218,9 @@ def _process_annotated_type(
     field_info: FieldInfo
     match annotation_metadata:
         case str():
-            field_info = cast(FieldInfo, Field(default=default, description=annotation_metadata))
+            field_info = cast("FieldInfo", Field(default=default, description=annotation_metadata))
         case FieldInfo():
-            field_info = cast(FieldInfo, annotation_metadata)
+            field_info = cast("FieldInfo", annotation_metadata)
             field_info.default = default
         case _:
             raise ValueError(f"Unsupported annotation metadata: {annotation_metadata}")
@@ -236,7 +236,7 @@ def _process_regular_type(
     """Process regular type annotations without metadata."""
     param_type = param_annotation
     description = _form_param_description(param_name, func_name)
-    field_info = cast(FieldInfo, Field(default=default, description=description))
+    field_info = cast("FieldInfo", Field(default=default, description=description))
     return param_type, field_info
 
 
--- a/portia/tool_decorator.py
+++ b/portia/tool_decorator.py
@@ -3,20 +3,19 @@
 from __future__ import annotations
 
 import inspect
-from collections.abc import Awaitable, Callable
-from typing import Annotated, Any, TypeVar, get_args, get_origin, get_type_hints
-
+from typing import Any, TypeVar, get_origin, get_args, get_type_hints
 from pydantic import BaseModel, Field, create_model
 from pydantic.fields import FieldInfo
-
 from portia.tool import Tool, ToolRunContext
+from typing import Annotated
+from collections.abc import Awaitable, Callable
 
 # Type variables for the decorator
 P = inspect.Parameter
 T = TypeVar("T")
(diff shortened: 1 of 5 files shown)


jingkaihe:
fwiw looks like the eval gha workflow is failing because the checkout step on gha workflow doesn't support checking out from forked branch - https://github.com/portiaAI/portia-sdk-python/blob/main/.github/workflows/evals.yml#L38

Nathanjp91 (commented):
Thanks for the contribution. Really Like the approach. I'll be dogfooding it a bit more on monday before full approval but looks good as long as all the checks pass (aside from evals which I'll be attempting to resolve)

There's some extensions I would like but they're not listed in the relevant issue so won't be blocking. For instance, we now have structured_output_schema for coercion of tool outputs to pydantic models, but I think the only way to approach that here would be to put that as an argument in to the tool decorator, like so

```python
class MyOutputModel(Basemodel):
    ...

@tool(MyOutputModel) # register structured_output_schema
def my_method(...):
    ...
```

Nothing blocking there though as it's not part of the listed issue, but this interface might get changed later to include it.

Sorry the checks are a huge faff, this is the first outside contribution so none of them have been tested in anger against forks rather than internal PRs.

portia/tool_decorator.py
```diff
@@ -0,0 +1,261 @@
...
+    args_schema = _create_args_schema(sig, func_name, fn)
+
+    # Determine output schema from return type
+    output_schema = _create_output_schema(type_hints, func_name)
+
+    # Create the Tool class dynamically
+    def make_run_method() -> Callable:
+        def run(self: Tool[T], ctx: ToolRunContext, **kwargs: Any) -> T:  # noqa: ARG001
+            """Execute the original function."""
+            # Filter out 'ctx' parameter if the function doesn't expect it
```
emmaportia, line 80: Love this – this is a very nice way of handling this, thank you!
Will you also follow up to this PR with updates to the docs?
I think our original issue description didn’t include doc updates so happy to do that for 40 extra

portia/tool_decorator.py
```diff
@@ -0,0 +1,261 @@
...
+    Raises:
+        ValueError: If the function has invalid signature or return type
+
+    """
+    # Validate function
+    _validate_function(fn)
+
+    # Extract function metadata
+    func_name = fn.__name__
+    description = (fn.__doc__ or "").strip()
```
emmaportia, line 60: I’m pretty sure both our tool ids and names might have specific characters that cause validation to fail. (e.g : from memory) – could we add some tests for that and also make sure that we’re appropriately removing / escaping those characters?

portia/tool_decorator.py
```diff
@@ -0,0 +1,261 @@
...
+
+def _extract_type_and_field_info(
+    param_annotation: type | object,
+    param: inspect.Parameter,
+    param_name: str,
+    func_name: str,
+) -> tuple[type | object, FieldInfo]:
+    """Extract type and field information from parameter annotation.
+
+    Supports:
```
emmaportia, line 180: This is great. OOC (and not necessary for this PR), but what are your thoughts on supporting this within docstring function annotations?

emmaportia (commented):
Thanks! This is great, some minor nits!

portia/tool_decorator.py
jingkaihe: There are tool ids duplication check in place, can't find any about the name afaik. Although there are max characters checks on the description. I added some test cases for them here - https://github.com/portiaAI/portia-sdk-python/pull/519/commits/1da9b84321dfc0ad79dbc72c3857c02bd706854f

portia/tool_decorator.py
jingkaihe: Can do. Do you want it to be done in the doc string, or a separate doc is needed?

portia/tool_decorator.py
emmaportia: yeah true true, i think that's fine

PR #535, opened 2025-06-19
clarifications enabled feature flag removal

Nathanjp91:
# Description

Ticket Link: N/A 

## Type of change

(select all that apply)

- [ ] Bug fix 
- [ ] New feature 
- [x] Breaking change 
- [x] Refactor
- [ ] Documentation update

## Screenshots

## Changelog


portia/config.py
```diff
@@ -468,6 +466,10 @@ class Config(BaseModel):
         default={},
         description="A dictionary of feature flags for the SDK.",
     )
+    clarifications_enabled: bool = Field(
+        default=False,
+        description="Whether to enable clarifications for the execution agent.",
```
OmarEl-Mohandes, line 471: We can be more specific here. something like `This is to enable ClarificationTool in the execution agent.`. Since one_shot can still return clarifications from tool regardless of this.

OmarEl-Mohandes (approved):

PR #594, opened 2025-07-18
Update tool_registry.py

amelia-mich:
Added a with_tool_description_regex method to allow for regexs

# Description

Added a with_tool_description_regex method to allow for regexs

## Type of change

(select all that apply)

- [ ] New feature

### DIFF
--- a/portia/tool_registry.py
+++ b/portia/tool_registry.py
@@ -203,6 +203,26 @@ class ToolRegistry:
 
         """
         return ToolRegistry({tool.id: tool for tool in self._tools.values() if predicate(tool)})
+    def with_tool_description_regex(
+        self, regex: str, updated_description: str, *, overwrite: bool = False
+    ) -> ToolRegistry:
+        """Update tools that fit the regex expression with an extension or override of the tool description
+
+        Args:
+            regex (str): A regex expression to describe the tools to update.
+            updated_description (str): The tool description to update. If `overwrite` is False, this
+                will extend the existing tool description, otherwise, the entire tool description
+                will be updated.
+            overwrite (bool): Whether to update or extend the existing tool description.
+        """
+        try:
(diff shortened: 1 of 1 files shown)


portia/tool_registry.py
```diff
@@ -203,6 +203,26 @@ def filter_tools(self, predicate: Callable[[Tool], bool]) -> ToolRegistry:
 
         """
         return ToolRegistry({tool.id: tool for tool in self._tools.values() if predicate(tool)})
+    def with_tool_description_regex(
+        self, regex: str, updated_description: str, *, overwrite: bool = False
+    ) -> ToolRegistry:
+        """Update tools that fit the regex expression with an extension or override of the tool description
+
```
emmaportia, line 210: Probably worth a quick example here to explain how it can be used with mcp?

portia/tool_registry.py
```diff
@@ -203,6 +203,26 @@ def filter_tools(self, predicate: Callable[[Tool], bool]) -> ToolRegistry:
...
+        Args:
+            regex (str): A regex expression to describe the tools to update.
+            updated_description (str): The tool description to update. If `overwrite` is False, this
+                will extend the existing tool description, otherwise, the entire tool description
+                will be updated.
+            overwrite (bool): Whether to update or extend the existing tool description.
+        """
+        try:
+            filtered_tools = self.filter_tools(lambda tool: re.match(regex, tool.id) is not None)
+            for tool in filtered_tools:
```
emmaportia, line 220: you could do this more neatly with a list comprehension (and would be the pythonic way to do it), you could also do line 219 as well, i.e something like

[self.with_tool_description(...) for tool in self.filter_tools(...)]

emmaportia (commented):
Thanks!

emmaportia (commented):
Codes looking good! Just the tests before we can get this in :)

PR #655, opened 2025-08-14
Add llm tool default back

robbie-portia:
# Description

This adds the LLM tool default back (see slack for details) but tries to fetch it from the registry first. This is needed in order to make the tool stubbing work within steel thread.

## Type of change

(select all that apply)

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/portia/portia.py
+++ b/portia/portia.py
@@ -45,6 +45,7 @@ from portia.errors import (
     PlanError,
     PlanNotFoundError,
     SkipExecutionError,
+    ToolNotFoundError,
 )
 from portia.execution_agents.base_execution_agent import BaseExecutionAgent
 from portia.execution_agents.default_execution_agent import DefaultExecutionAgent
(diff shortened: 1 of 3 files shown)


robbie-portia:
Good point @Nathanjp91 - I've now made it so that the LLMTool is automatically added to the registry rather than doing it when getting the tool. I originally had it configurable via an optional parameter when you create the registry, but that caused issues when we create new registries internally by adding registries (we'd need to store whether we wanted to auto-add the llm_tool and pass it through, and handle clashes etc.) - I think it's easier to just always add the llm_tool. This isn't a change of behaviour from today and people can always remove it from the registry (though they'd need to be careful not to add any other registries).

robbie-portia:
I also realised that it's a bit weird to be auto-adding LLMTool to MCP registries. Chatted with @Nathanjp91 and we thought it was actually easiest to go back to the original method of just giving back LLMTool when trying to get_tool, even if it isn't in the registry.

Nathanjp91 (approved):

PR #743, opened 2025-08-29
Fix linked step inputs 

robbie-portia:
# Description

Previously, things worked with a plan input value of StepOutput, but not with string templating e.g. "...StepOutput(0)...". This was because we weren't handling the string templating / reference value lookups recursively (i.e each string template might have a reference, which itself might have a value that is a string with a reference etc.). Not sure I've explained that perfectly, so happy to zoom if easier, but this PR fixes the issue. It also tidies things up by removing ReferenceValue which I think wasn't needed - now doing `get_value()` on a reference just returns the raw value without a wrapper.

## Type of change

(select all that apply)

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog

### DIFF
--- a/.github/workflows/contrib.yml
+++ b/.github/workflows/contrib.yml
@@ -1,0 +1,19 @@
+name: Update contributors
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  contrib-readme-job:
+    runs-on: ubuntu-latest
+    name: A job to automate contrib in readme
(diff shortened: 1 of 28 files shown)


Nathanjp91 (approved):
Sorry, forgot to hit submit on this! LGTM

PR #750, opened 2025-09-03
Add ReactAgentStep

robbie-portia:
# Description

Hopefully it's all self-explanatory - see example_builder.py for how it works

## Type of change

(select all that apply)

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Refactor
- [ ] Documentation update

## Screenshots

## Changelog


portia/execution_agents/react_clarification_tool.py
```diff
@@ -0,0 +1,37 @@
...
+    description: str = (
+        "Raises a clarification with the user if you do not have enough information to proceed"
+    )
+    args_schema: type[BaseModel] = ReActClarificationToolSchema
+    output_schema: tuple[str, str] = ("InputClarification", "The clarification to raise")
+
+    def run(self, ctx: ToolRunContext, guidance: str) -> InputClarification:
+        """Run the ReActClarificationTool."""
+        return InputClarification(
+            argument_name="react_agent_clarification",
```
robbie-portia, line 33: Note: This is a mis-use of `InputClarification` (as it doesn't refer to an argument), but we do the same with `UserInputStep`

portia/open_source_tools/sql_tool.py
```diff
@@ -52,25 +52,25 @@ def _sqlite_authorizer(
     """
     # Allow read operations
     if action in (
-        sqlite3.SQLITE_READ,      # Reading data from tables
-        sqlite3.SQLITE_SELECT,    # SELECT statements
-    sqlite3.SQLITE_FUNCTION,  # Using functions in queries
-):
+        sqlite3.SQLITE_READ,  # Reading data from tables
```
robbie-portia, line 55: These are linting changes from pre-commit from a recent open-source contribution. I'm going to update our linting in Github so this can't happen. Same with tests/unit/open_source_tools/test_sql_tool.py

portia/builder/step_v2.py
```diff
@@ -472,6 +473,85 @@ def to_legacy_step(self, plan: PlanV2) -> Step:
         )
 
 
+class ReActAgentStep(StepV2):
+    """A step that uses a ReAct agent to complete a task."""
+
+    task: str = Field(description="The task to perform.")
+    tools: list[str] = Field(description="The tools to use.")
```
Nathanjp91, line 480: maybe not required now, but it would be nice to decouple this further from the Portia side of things. At the moment this kind of relies on a really round about method to get a tool react agent.
Instead of react agent pulling from run_data.portia.tool_registry, being able to just give it the tools directly, (but still allow str name matching against the tool_registry.

portia/portia.py
```diff
@@ -2638,6 +2638,25 @@ def _check_remaining_tool_readiness(
 
         return ready_clarifications
 
+    def resolve_tool_id_with_comma(self, tool_id: str) -> list[str]:
```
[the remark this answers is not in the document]
Nathanjp91: Could we just add a validator to Tool to ensure there's no comma in the name, then just always do tool_id.split(','), basically removing the first case and assuming any comma seperation is an agent listing tools?

portia/builder/plan_builder_v2.py
```diff
@@ -283,6 +284,47 @@ def single_tool_agent_step(
...
+    def react_agent_step(
+        self,
+        *,
+        task: str,
+        tools: list[str] | None = None,
+        inputs: list[Any] | None = None,
+        output_schema: type[BaseModel] | None = None,
+        step_name: str | None = None,
+        allow_agent_clarifications: bool = False,
+        tool_call_limit: int = 25,
```
Nathanjp91, line 296: Think this can come later, but it would be nice to also have a guardrails function. Something that is part of the langgraph node for the React Agent that is called after each iteration. Default implementation to just return, but something the user can provide that can look at the outputs of the agent and decide to exit early. 

The other solution, which might be more extensible, but requires more to implement, is to expose a compositional factory argument here with default to our agent, that allows them to provide their own guardrail'ed react agent to this step. Something like
```python
def react_agent_step(
    ...
    factory: type[ReActAgentStep] = ReActAgentStep
)
    ...
    self.plan.steps.append(factory(...))
    return self
```

This way can implement your own version of a ReAct agent without having to write a new PBv2 function for it. So long as it conforms to the interface.

portia/execution_agents/react_agent.py
```diff
@@ -0,0 +1,444 @@
...
+            ExecutionAgentUsageTelemetryEvent(
+                agent_type="react",
+                model=str(self.run_data.config.get_execution_model()),
+                sync=False,
+                tool_id=",".join([tool.id for tool in self.tools]),
+            )
+        )
+
+        tool_run_ctx = self.run_data.get_tool_run_ctx()
+        # We use the planning model rather than the execution model here as
```
Nathanjp91, line 307: I'd be wary of this actually. A lot of steps in a ReAct agent can be quite simple, and our planning models are typically thinking models which often show worse performance on simple tasks. It would be interesting to play around with a ReAct agent setup that worked out the iterative tasks it had to do alongside a task complexity score and then routed to different models based on complexity. Beyond this implementation though I think.

Nathanjp91 (approved):
LGTM: Nothing blocking I think. Just lots of directional comments.

portia/builder/step_v2.py
robbie-portia: Yeah, that's definitely fair - I'd definitely like to tidy this up in future

portia/execution_agents/react_agent.py
robbie-portia: Yeah, I definitely wasn't sure on this - I was thinking that we'd generally want a thinking model for the ReAct agent (as they are best for the reasoning part, which is where the model is used). Will be interesting to see what people think during dogfooding.

